PR #32, opened 2024-05-20
ci: attempt to check in the CI build of the dist dir.

chirino:
I think this should end up creating `static-${ref}` tags like `static-main` tags in the repo which will contain the built dist dir.  If this does work, maybe we can change the crate to just use the existing dist files instead of doing a build?

### DIFF
--- a/.github/workflows/tag-dist.yaml
+++ b/.github/workflows/tag-dist.yaml
@@ -1,0 +1,33 @@
+name: Tag Dist Build
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+
+jobs:
+  tag-dist:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - name: Install
+        run: npm clean-install --ignore-scripts
+      - name: Build
+        run: npm run build
+      - name: Commit and Tag the static dist dir
+        shell: bash
+        run: |
+          git config --global user.name "${GITHUB_ACTOR}"
+          git config --global user.email "${GITHUB_ACTOR}@users.noreply.${INPUT_ORGANIZATION_DOMAIN}"
+          git add -f client/dist
+          git commit -m "Checkin dist dir"
+          git fetch --tags
+          git tag static-${{ github.ref_name }} --force
+          git push origin static-${{ github.ref_name }} --force
+
--- a/crate/build.rs
+++ b/crate/build.rs
@@ -1,7 +1,8 @@
-use static_files::resource_dir;
 use std::path::Path;
 use std::process::{Command, ExitStatus};
 use std::{fs, io};
+
+use static_files::resource_dir;
 
 static UI_DIR: &str = "../";
 static UI_DIR_SRC: &str = "../src";
@@ -18,15 +19,9 @@ fn main() {
 
     println!("cargo:rerun-if-changed={}", UI_DIR_SRC);
 
-    let build_ui_status = install_ui_deps()
-        .and_then(|_| build_ui())
-        .and_then(|_| copy_dir_all(UI_DIST_DIR, STATIC_DIR));
+    build_ui().expect("Error while building UI");
 
-    match build_ui_status {
-        Ok(_) => println!("UI built successfully"),
-        Err(_) => panic!("Error while building UI"),
-    }
-
+    copy_dir_all(UI_DIST_DIR, STATIC_DIR).expect("Failed to copy UI files");
     resource_dir("./target/generated").build().unwrap();
 }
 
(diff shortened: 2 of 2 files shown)


helio-frota (approved):

carlosthe19916:
@chirino thanks for PR, I really appreciate it.
- Do you mind creating also a PR into the trustify repository to have the full picture?
- I personally do not like the idea of creating tags for each commit into this repository. We can publish the tags in another repository and avoid polluting this one?

In general, I'm convinced (after all discussion internally) we all are not clear if we are doing this for "dev mode" or "prod mode". 

- My personal interpretation of "dev mode" would be:
  - `cargo run` for starting the backend and `npm run start:dev` for starting the UI
- For Prod mode, we execute only the final binary generated by any kind of CI

What scenario do you think we are solving "dev mode"? or "prod mode"? I am not trying to block your PR but I would like to solve the problem and internal discussions sooner rather than later :)

carlosthe19916 (changes requested):
Comment above

chirino:
I agree with using another repo long term.  But for the short term, it might be easier to use this one to validate the idea.

I think the dev/prod issue is that you have different types of dev mode.  Some devs only want to do dev on the trusttify server, not the UI.  This change will make life easier for those, as they won't need to build the UI bits every time they run the rust build.

helio-frota (approved):

carlosthe19916 (approved):
Approving the PR as a vehicle to validate the idea but not the final solution

chirino:
the latest attempt to check in the tag via CI failed with:

```
 create mode 100644 client/dist/monaco/ts.worker.js
 create mode 100644 client/dist/monaco/ts.worker.js.LICENSE.txt
 create mode 100644 client/dist/monaco/ts.worker.js.map
From https://github.com/trustification/trustify-ui
 * [new branch]      main       -> origin/main
remote: Permission to trustification/trustify-ui.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/trustification/trustify-ui/': The requested URL returned error: 403
Error: Process completed with exit code 128.
```

So I think we need to update the github setting for this repo as described in https://stackoverflow.com/questions/72851548/permission-denied-to-github-actionsbot

carlosthe19916 (changes requested):
- @chirino I have verified and this repository has already the correct permissions "red and write" for gh workflows
- The reason of the 403 error is that this PR comes from your forked repository and due to good security reasons it is not allowed to generate tags on the target repository through PRs. So your ` tag-dist` job should never run against PRs but only on commits done in the main branch (this is change to do in your PR). The following definition is what needs to be changed:

```
name: CI

on:
  push:
    branches:
      - main
      - "release-*"
  pull_request:
    branches:
      - main
      - "release-*"
```

- I didn't see any counterpart PR on https://github.com/trustification/trustify for using the changes done in this repository, is the expectation that no changes on the trustify repository will be needed? If not, I don't think it is a good idea to merge this PR without being sure it won't break things on the trustify repository. The trustify repository is already using the crate within the trustify-ui repository here https://github.com/trustification/trustify/blob/main/Cargo.toml#L130 and if you merge this PR without an immediate  change to the trustify repo it will break the trustify repository intermediately. Let's avoid that because it will cause troubles to everyone working in the trustify repository.

I'd like us both to take a step back and validate ideas correctly before merging things on the main branch. What about this:

1. Using your forked repository execute the commands defined in this PR (either in a CI action or even manually in your terminal), then publish the tag in your forked repository.
2. Create a PR in the Trustify repository and make use of your forked-repository/tag here https://github.com/trustification/trustify/blob/main/Cargo.toml#L130 
3. We can see how it goes and truly validate the idea without polluting neither trustify nor trustify-ui.

What do you think?

chirino:
Will do.

chirino:
Tried with this patch against trustify: 

```patch
diff --git a/Cargo.toml b/Cargo.toml
index 8c2d1cd..a18b1c9 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,7 +124,7 @@ trustify-module-ui = { path = "modules/ui", default-features = false }
 trustify-module-advisory = { path = "modules/advisory" }
 trustify-module-vulnerability = { path = "modules/vulnerability" }
 trustify-server = { path = "server", default-features = false }
-trustify-ui = { git = "https://github.com/trustification/trustify-ui.git" }
+trustify-ui = { git = "https://github.com/chirino/trustify-ui.git", tag = "static-main" }
 ```

It built fine and UI loaded.

**Comment by carlosthe19916** [comment]
Thanks @chirino I think now I have a good understanding of your strategy in this PR. Sounds good to me
- After https://github.com/trustification/trustify/pull/319 is approved we can also merge this one.
- Later on the publishing of the tags must be done in another repository and not in this one
- The github action that generates the tags on each PR should not run on each PR but only on new commits on the main branch

**Review by carlosthe19916 — APPROVED** [review]

File: .github/workflows/ci-actions.yaml [hunk 1]
```diff
@@ -39,4 +39,29 @@ jobs:
...
+      - name: Commit and Tag the static dist dir
+        shell: bash
+        run: |
+          git config --global user.name "${GITHUB_ACTOR}"
+          git config --global user.email "${GITHUB_ACTOR}@users.noreply.${INPUT_ORGANIZATION_DOMAIN}"
+          git add -f client/dist
+          git commit -m "Checkin dist dir"
+          git fetch --tags
+          git tag static-${{ github.ref_name }} --force
+          git push origin static-${{ github.ref_name }} --force
```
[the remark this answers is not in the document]
chirino: ah your right, I'll have to move this to different file then.

chirino:
fixed and squashed

.github/workflows/tag-dist.yaml
```diff
@@ -0,0 +1,34 @@
...
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+
+jobs:
+  tag-dist:
+    runs-on: ubuntu-latest
+    needs: ["ci"]
```
carlosthe19916, line 13: This might fail. besides it, looks fine, feel free to merge it once this minor thing is fixed

chirino: good catch.  fixed.

chirino:
I can't merge. :(

PR #47, opened 2024-06-07
Avoid to use stack when not needed

gildub:
Removes unnecessary call on the stack.
Definitely not a big deal and although it looks more tidy it's not just about cosmetic.


carlosthe19916:
I personally prefer the "return" statements as it makes easier to read the code, but I have no strong opinion in this case, so let's merge it :) . Thanks Gilles!

carlosthe19916 (approved):

PR #67, opened 2024-06-20
task: refactor model namings and query locations

carlosthe19916:
No functional change, only refactoring code:

- Rename some models to be more verbose
- Move some query functions to their appropriate files
- Order some rest functions within the same file


gildub (approved):
LGTM

PR #70, opened 2024-06-24
Fix: UTC time to active importers

gildub:
Add /src/app/dayjs.ts with common options
Add UTC time to importers 
Add `formatDateTime` function

Fix https://github.com/trustification/trustify-ui/issues/68

Screenshot from 2024-06-24 15-22-21

### DIFF
--- a/client/src/app/dayjs.ts
+++ b/client/src/app/dayjs.ts
@@ -3,12 +3,10 @@ import isSameOrBefore from "dayjs/plugin/isSameOrBefore";
 import utc from "dayjs/plugin/utc";
 import timezone from "dayjs/plugin/timezone";
 import customParseFormat from "dayjs/plugin/customParseFormat";
-import relativeTime from "dayjs/plugin/relativeTime";
-import localizedFormat from "dayjs/plugin/localizedFormat";
+import arraySupport from "dayjs/plugin/arraySupport";
 
 dayjs.extend(utc);
 dayjs.extend(timezone);
 dayjs.extend(customParseFormat);
 dayjs.extend(isSameOrBefore);
-dayjs.extend(relativeTime);
-dayjs.extend(localizedFormat);
+dayjs.extend(arraySupport);
--- a/client/src/app/dayjs.ts
+++ b/client/src/app/dayjs.ts
@@ -1,0 +1,14 @@
+import dayjs from "dayjs";
+import isSameOrBefore from "dayjs/plugin/isSameOrBefore";
+import utc from "dayjs/plugin/utc";
+import timezone from "dayjs/plugin/timezone";
+import customParseFormat from "dayjs/plugin/customParseFormat";
+import relativeTime from "dayjs/plugin/relativeTime";
+import localizedFormat from "dayjs/plugin/localizedFormat";
+
+dayjs.extend(utc);
+dayjs.extend(timezone);
+dayjs.extend(customParseFormat);
+dayjs.extend(isSameOrBefore);
+dayjs.extend(relativeTime);
+dayjs.extend(localizedFormat);
--- a/client/src/app/pages/advisory-details/vulnerabilities.tsx
+++ b/client/src/app/pages/advisory-details/vulnerabilities.tsx
@@ -130,14 +130,10 @@ export const Vulnerabilities: React.FC<VulnerabilitiesProps> = ({
                       {item.non_normative.title}
                     </Td>
                     <Td width={10} {...getTdProps({ columnKey: "discovered" })}>
-                      {!!item.non_normative.discovered
-                        ? formatDate(item.non_normative.discovered)
-                        : null}
+                      {formatDate(item.non_normative.discovered)}
                     </Td>
                     <Td width={10} {...getTdProps({ columnKey: "released" })}>
-                      {!!item.non_normative.released
-                        ? formatDate(item.non_normative.released)
-                        : null}
+                      {formatDate(item.non_normative.released)}
                     </Td>
                     <Td width={15} {...getTdProps({ columnKey: "severity" })}>
                       <SeverityShieldAndText value={item.severity} />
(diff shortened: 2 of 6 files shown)


helio-frota (approved):
lgtm

client/src/app/pages/importer-list/importer-list.tsx
```diff
@@ -250,14 +251,21 @@ export const ImporterList: React.FC = () => {
                           modifier="truncate"
                           {...getTdProps({ columnKey: "start" })}
                         >
-                          {formatDate(item.report?.startDate)}
+                          {item.state &&
+                            configValues?.disabled == false &&
+                            dayjs(item.report?.startDate)
+                              .utc()
+                              .format("YYYY-MM-DD HH:mm:ss")}
```
carlosthe19916, line 258: Just an idea:
-  The current code in the main branch has a function `formatDate()` that can be reused everywhere.
- what about having something like `formatDateTime` so we can reuse it too?

carlosthe19916 (changes requested):
Minor comments, please take a look

client/src/app/dayjs.ts
```diff
@@ -0,0 +1,14 @@
...
+import customParseFormat from "dayjs/plugin/customParseFormat";
+import relativeTime from "dayjs/plugin/relativeTime";
+import localizedFormat from "dayjs/plugin/localizedFormat";
+
+dayjs.extend(utc);
+dayjs.extend(timezone);
+dayjs.extend(customParseFormat);
+dayjs.extend(isSameOrBefore);
+dayjs.extend(relativeTime);
+dayjs.extend(localizedFormat);
```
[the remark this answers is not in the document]
gildub: @carlosthe19916, yes indeed, I missed the import file, I fixed it, thanks

client/src/app/pages/importer-list/importer-list.tsx
gildub: @carlosthe19916, 

Well I suppose this a question of reuse vs performance.
I'm not against a format function here but what's do we gain as the call to dayjs format is straightforward here.

carlosthe19916: Having formats in a single place will help us to change the formats in a single shot when needed. So far this is the only place where we use this format but in the case of `formatDate` if UX decides the format needs to change We will need to only change that function and the whole app will have the desired format rather than changing `.format(ABCAD)` in every single file

carlosthe19916:
@helio-frota your report is not accurate. Pull the latest changes of the main branch

helio-frota:
@carlosthe19916 that's correct sorry for the noise, the function is being used and using `dayjs`.

in this case I'm +1 having a `formatDateTime`

helio-frota (changes requested):
There is a function inside `client/src/app/utils/utils.ts` that makes use of dayjs already.

export const formatDate = (value?: string) => {
  return value ? dayjs(value).format(RENDER_DATE_FORMAT) : undefined;
};

so I'm +1 to @carlosthe19916 suggestion

helio-frota (approved):

client/src/app/pages/advisory-details/vulnerabilities.tsx
```diff
@@ -137,10 +130,14 @@ export const Vulnerabilities: React.FC<VulnerabilitiesProps> = ({
                       {item.non_normative.title}
                     </Td>
                     <Td width={10} {...getTdProps({ columnKey: "discovered" })}>
-                      {formatDate(item.non_normative.discovered)}
+                      {!!item.non_normative.discovered
+                        ? formatDate(item.non_normative.discovered)
+                        : null}
```
carlosthe19916, line 135: why the unnecessary change of this block of code? what problem are we solving with this?

gildub: @carlosthe19916, It's because I've made changes to https://github.com/trustification/trustify-ui/pull/70/files#diff-93d1e679f3a44569699332c49a4e0cf4e30ada958d50d897b3764cf2e20dd6f7R37-R39 where it makes more sense to not return `undefined` and therefore check for such case before calling `formatDate`.

carlosthe19916: - Sorry but this change is replacing a simple line of code with a complex conditional longer code. This is is a step back in my opinion.
- What specific problem is this change solving that the previous code was not covering?

gildub: @carlosthe19916, 

`formatDate(item.non_normative.discovered)` what happens when formatDate returns undefined ?
React might break in such case and what should be done is to trap such case and return null.

carlosthe19916: - `formatDate` is already returning undefined in multiple cases. 
```shell
const myVar = formatDate(undefined); // Then myVar will be undefined

<>
{myVar}
</>
```

How is the previous code breaking reactjs? Can you give me a real concrete example and not suppositions?

-  I think the case is simple: Case1 vs Case2, let's explain why:

```shell
{!!item.non_normative.discovered
                        ? formatDate(item.non_normative.discovered)
                        : null}
```

is better than 

```
{formatDate(item.non_normative.released)}
```

Why if `item.non_normative.released` is undefined it will break the code?

- Couldn't we just do the following?:
```shell
export const formatDate = (value?: string) => {
  return value ? dayjs(value).format(RENDER_DATE_FORMAT) : null;
};
```

we are returning null as you want.

gildub: @carlosthe19916 , sorry for the confusion, what I mean by breaking the code is not reactjs friendly when the code renders {undefined}.
Well it seems that React 18 handles undefined better. But that shouldn't really happen anyway : 
https://blog.saeloun.com/2022/04/14/react-18-allows-components-to-render-undfined/

Let's use your latest example and make sure formatDate and formatDateTime return null if undefined.

carlosthe19916 (approved):

PR #76, opened 2024-06-26
advisory/vulnerability Add cwe

carlosthe19916:


client/src/app/pages/advisory-details/vulnerabilities.tsx
```diff
@@ -146,8 +146,7 @@ export const Vulnerabilities: React.FC<VulnerabilitiesProps> = ({
                       <SeverityShieldAndText value={item.severity} />
                     </Td>
                     <Td width={10} {...getTdProps({ columnKey: "cwe" })}>
-                      {/* {item.cwe} */}
-                      <p style={{ color: "red" }}>issue-281</p>
+                      {item.non_normative.cwe}
```
gildub, line 149: Could you please test for undefined value as cwe being optional it could be undefined and be rendered as such ?

gildub (changes requested):

client/src/app/pages/advisory-details/vulnerabilities.tsx
carlosthe19916: @gildub I think now I understand where you are coming from. We are using ReactJS 18 and it is safe to render undefined values.

gildub (approved):
LGTM

PR #78, opened 2024-06-26
Package list: Set qualifiers on main row

carlosthe19916:
fixes https://github.com/trustification/trustify-ui/issues/64

### DIFF
--- a/client/src/app/pages/package-list/package-list.tsx
+++ b/client/src/app/pages/package-list/package-list.tsx
@@ -56,7 +56,7 @@ export const PackageList: React.FC = () => {
       namespace: "Namespace",
       version: "Version",
       type: "Type",
-      path: "Path",
+      qualifiers: "Qualifiers",
       cve: "CVEs",
     },
     isPaginationEnabled: true,
@@ -157,7 +157,7 @@ export const PackageList: React.FC = () => {
                   <Th {...getThProps({ columnKey: "namespace" })} />
                   <Th {...getThProps({ columnKey: "version" })} />
                   <Th {...getThProps({ columnKey: "type" })} />
-                  <Th {...getThProps({ columnKey: "path" })} />
+                  <Th {...getThProps({ columnKey: "qualifiers" })} />
                   <Th {...getThProps({ columnKey: "cve" })} />
                 </TableHeaderContentWithControls>
               </Tr>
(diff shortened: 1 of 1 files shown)


gildub (changes requested):
All optional field can be undefined therefore we need to "trap" them so we don't render undefined, please see inline example.

Also it's recommended not to use && for equivalent rendering reasons.
It's very intuitive to simple use `condition && function or Component` but that could lead to trouble when the first part of the condition returns undefined.

client/src/app/pages/package-list/package-list.tsx
```diff
@@ -204,13 +206,12 @@ export const PackageList: React.FC = () => {
...
                         </Td>
                         <Td
-                          width={10}
-                          modifier="truncate"
-                          {...getTdProps({ columnKey: "path" })}
+                          width={30}
+                          {...getTdProps({ columnKey: "qualifiers" })}
                         >
-                          {item.decomposedPurl?.path}
+                          {item.decomposedPurl?.qualifiers && <PackageQualifiers value={item.decomposedPurl?.qualifiers} />}
```
gildub, line 212: It might be a matter of style but I think we need to be safer here and by doing so try to be consistent by using `condition ? <render value> : null` kind of approach.

gildub (changes requested):

carlosthe19916:
@gildub I think we will discuss the same topic in multiple PRs/Issues xD

> All optional field can be undefined therefore we need to "trap" them so we don't render undefined, please see inline example.

What do you mean by rendering undefined? are you suggesting that if I have the following code:

```shell
var myVar=undefined;
var myComponent = <>{myVar}</>
```

Then myComponent will render the words `undefined` in the screen? is that what you are suggesting? If so, that is not the case, neither null nor undefined values are ever rendered in the HTML DOM

> It's very intuitive to simple use condition && function or Component but that could lead to trouble when the first part of the condition returns undefined.

Let's discuss a concrete case and not a generic/non-existen case. In which case the following code will case us troubles and lead us into a "trap"?

```
<>
{item.decomposedPurl?.qualifiers && <PackageQualifiers value={item.decomposedPurl?.qualifiers} />}
</>
```

To be more specific: How will we fall into a trap if `item.decomposedPurl?.qualifiers` is undefined?

gildub:
@carlosthe19916, my understanding might be obsolete, before Reactjs 18 undefined value was an issue. Since React 18 this seems not the case anymore. 

That is a generic question as how to handle such cases. 

I understand that dealing with a undefined value inside a React (v18) component ({undefined}) is not an issue as it won't break or even create a warning. Therefore are we ok to have undefined values surfacing or do we want to try to catch them as much as possible, that's the generic question.

helio-frota:
@gildub 

> Since React 18 this seems not the case anymore.

according to the react maintainer the check needs to be delegated to the lint 

```
Linting is better for this use case
At runtime, you can’t tell the difference between forgetting to return and explicitly returning undefined, but this is something linting can check for because it operates on the code syntax as it’s written instead of how it’s executed.

Linting is also able to give more specific messages for why you may have made an accident, such as “this function is missing a return statement” ([consistent-return](https://eslint.org/docs/2.0.0/rules/consistent-return)), or “this expression was unused” ([no-unused-expressions](http://no-unused-expressions/)). More specific messages with code frames pointing you to the source of the issue will help you find the issue quicker than a generic run time error.
```

carlosthe19916:
The fact is that we are using ReactJS 18 already.

>  Therefore are we ok to have undefined values surfacing or do we want to try to catch them as much as possible, that's the generic question.

I think, let's catch them when it makes sense. I don't think it makes sense in this specific PR. The fact that ReactJS 18 allows us to have undefined within components should allow us to have verbose and simple lines of code without complicating it when not needed

gildub (approved):
LGTM

PR #80, opened 2024-06-26
fix: do not include empty filter values

carlosthe19916:
Fixes https://github.com/trustification/trustify-ui/issues/74


gildub (approved):
LGTM

PR #82, opened 2024-06-27
feat: add sbom described_by

carlosthe19916:
Based on https://github.com/trustification/trustify/issues/284#issuecomment-2193993777

image

I don't really think this is the way of showing data, but at this point I have no clue how to render the `described_by` field. So I am just rendering it to have something to show and then on a future iteration let UX and PM decide how useful/correct the `described_by` data is  and how to render it.

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -102,11 +102,13 @@ export interface PackageWithinSBOM {
 
 export interface SBOM {
   id: string;
-  type: "CycloneDX" | "SPDX";
   name: string;
-  version: string;
   authors: string[];
   published: string;
+  described_by?: {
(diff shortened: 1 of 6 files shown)


gildub:
@carlosthe19916, I got the impression we need to wait for UXD/PM to decide what to expect and avoid leaving a trail of commented out lines of code. In the latter case we must use more explicit comments such as "// TODO - Blah-blah .." so we can track those.

gildub (commented):
Please see inline.

carlosthe19916:
@gildub thanks. I have remove all comments and added a TODO comment

gildub (approved):
LGTM

PR #83, opened 2024-06-27
feat: Packages within Vulnerabilities detail page

carlosthe19916:
Based on https://github.com/trustification/trustify/issues/462

We are supposed to see data in the table below 

image

The current problem is that neither me nor the backend guys were able to make the `advisories.statuses.affected` data appear.

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -33,9 +33,11 @@ export interface HubPaginatedResult<T> {
   params: HubRequestParams;
 }
 
-// Base
+// Advisories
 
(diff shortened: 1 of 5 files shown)


carlosthe19916:
@gildub 
With the help of the backend guys I was able to find a way to have data to test this.
- Have a clean instance of trustify and then upload the following advisory [RUSTSEC-2021-0079.json](https://github.com/user-attachments/files/16026886/RUSTSEC-2021-0079.json)
- The go to the vulnerability details page as the image below

image

gildub (approved):
LGTM

PR #86, opened 2024-06-28
fix: Fix stale index page by reporting the timestamp.

ctron:
Instead of using zero as the timestamp, use the actual timestamp. This should fix the issue of a stale page, when updating the UI. In that case the index page is not fetched, and it tries to load the old resources with the (no longer existing) cache busting markers. Resulting in a 404, and resulting in a blank page.

Also, report error through result

Instead of rendering one error into the page and reporting the others through a panic, report all through a result.

### DIFF
--- a/crate/Cargo.toml
+++ b/crate/Cargo.toml
@@ -4,11 +4,12 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-static-files = "0.2.1"
-serde_json = "1.0.117"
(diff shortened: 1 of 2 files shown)


ctron:
After this is merged, we also need to update the usage of `trustify_ui` in the `trustify` repository.

bobmcwhirter (approved):
I have no idea what this means, but I trust you.

PR #91, opened 2024-07-02
Update package url path changes

carlosthe19916:
Changes done in https://github.com/trustification/trustify/pull/468


carlosthe19916:
A second update was done here https://github.com/trustification/trustify/pull/495 therefore my last commit changes again the base path

PR #101, opened 2024-07-05
fix: correct field name

ctron:
This will also be fixed on the backend side.

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -185,14 +185,14 @@ export interface ImporterReport {
   report: {
     startDate: string;
     endDate: string;
-    numerOfItems: string;
+    numberOfItems: string;
   };
 }
 
 export interface ImporterReportDetails {
   startDate: string;
   endDate: string;
-  numerOfItems: string;
+  numberOfItems: string;
 }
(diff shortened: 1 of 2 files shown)


ctron:
Ok, looks like it takes a bit more. @carlosthe19916 maybe you can finish this up, or start from scratch with this is a pointer.

client/src/app/api/models.ts
```diff
@@ -185,14 +185,14 @@ export interface ImporterReport {
   report: {
```
gildub, line 185: While at it, why not use interface definition for report instead ?
```ts
export interface ImporterReport {
  id: string;
  error?: string;
  report: ImporterReportDetails;
}
```

gildub (changes requested):
The field name needs to be changed in `./src/app/pages/importer-list/importer-list.tsx`./src/app/pages/importer-list/importer-list.tsx too.

carlosthe19916:
@gildub I pushed a commit to this branch to fix the missing change on ./src/app/pages/importer-list/importer-list.tsx .

carlosthe19916 (approved):

PR #112, opened 2024-07-17
fix: Importer reports: use API order

gildub:
Remove unnecessary sorting importers reports when managing the items in `useLocalTableControls`.
Although sorting out reports using only the `startDate` should be enough and using both start and end dates are likely to break the original order therefore as the API is already providing the reports in timely fashioned order then there is no need to reorder the items.

Before: 
image

After: 
image

Fix https://github.com/trustification/trustify-ui/issues/104

### DIFF
--- a/client/src/app/pages/importer-list/importer-list.tsx
+++ b/client/src/app/pages/importer-list/importer-list.tsx
@@ -539,10 +539,6 @@ export const ImporterExpandedArea: React.FC<ImporterExpandedAreaProps> = ({
     initialItemsPerPage: 5,
     isSortEnabled: true,
     sortableColumns: ["startDate", "endDate"],
-    getSortValues: (report) => ({
-      startDate: dayjs(report.report.startDate).millisecond(),
-      endDate: dayjs(report.report.startDate).millisecond(),
-    }),
     isFilterEnabled: false,
     isExpansionEnabled: false,
   });
--- a/client/src/app/pages/importer-list/importer-list.tsx
+++ b/client/src/app/pages/importer-list/importer-list.tsx
@@ -541,7 +541,7 @@ export const ImporterExpandedArea: React.FC<ImporterExpandedAreaProps> = ({
     sortableColumns: ["startDate", "endDate"],
     getSortValues: (report) => ({
       startDate: dayjs(report.report.startDate).valueOf(),
-      endDate: dayjs(report.report.startDate).valueOf(),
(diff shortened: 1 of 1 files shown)


carlosthe19916 (changes requested):
@gildub thanks for pointing this issue.
I think removing the `getSortValues` might not be a complete solution. Even if we remove it and rely on the backend the user still can click on the columns and try to sort the tables on the client side.

I have found that the reason the current sorting is not working is because I committed the mistake of using `dayjs(DATE).millisecond()`. 
- The documentation here https://day.js.org/docs/en/durations/milliseconds mentions that millisecond is a number between 0 and 999 but I wrongly assumed that the function `millisecond` would return the equivalent of :

```javascript
new Date().getTime();
```

- What about changing the current to something like:
```javascript
getSortValues: (report) => ({
  startDate: new Date(report.report.startDate).getTime(),
  endDate: new Date(report.report.endDate).getTime()
}),
```

Or alternatively we can use the `valueOf` function of dayjs https://day.js.org/docs/en/display/unix-timestamp-milliseconds
 
```javascript
getSortValues: (report) => ({
      startDate: dayjs(report.report.startDate).valueOf(),
      endDate: dayjs(report.report.startDate).valueOf(),
    }),
```

gildub:
@carlosthe19916, thanks for investigating the real issue with the sort values.
I used your latest suggestion.

client/src/app/pages/importer-list/importer-list.tsx
```diff
@@ -540,8 +540,8 @@ export const ImporterExpandedArea: React.FC<ImporterExpandedAreaProps> = ({
     isSortEnabled: true,
     sortableColumns: ["startDate", "endDate"],
     getSortValues: (report) => ({
-      startDate: dayjs(report.report.startDate).millisecond(),
-      endDate: dayjs(report.report.startDate).millisecond(),
+      startDate: dayjs(report.report.startDate).valueOf(),
+      endDate: dayjs(report.report.startDate).valueOf(),
```
carlosthe19916, line 544:
```suggestion
      endDate: dayjs(report.report.endDate).valueOf(),
```

carlosthe19916: This is my own mistake. I used `startDate` in both lines. So it would be a good chance to correct my mistake

carlosthe19916 (approved):

PR #118, opened 2024-07-19
Avoid nesting objects definition

gildub:
To make the code easier to understand lets avoid nesting object definitions.

Using a non primitive type (or through an interface) helps for code clarity and not necessarily only when that nested definition is reused.
In both cases a meaningful name is desirable as much as possible (but not always an easy find).

Also replace `StatusType` type to `SbomStatus` because it offers a more description of the object and avoid overloading the name with its type.


carlosthe19916 (approved):
Thanks Gilles! LGTM!

PR #127, opened 2024-07-30
fix: remove "view source" in modal, from advisory list

carlosthe19916:
#### Problem
Rendering the source of Advisory files when they are big frozen the browser. Watch the second 33 of the following recording.

We can verify it uploading this CSAF file (4MB). 
[CVE-2024-24783.json](https://github.com/user-attachments/files/16430134/CVE-2024-24783.json)

https://github.com/user-attachments/assets/310925d8-7fc4-4dc0-b0a7-94bdbbf6e030

- Removing the "Source viewer" from the advisory list will avoid this issues.
- Surprisingly the "Source viewer" works fine inside the tab (so leaving it there unless we find out it is a bad idea)

image

### DIFF
--- a/client/src/app/pages/advisory-list/advisory-list.tsx
+++ b/client/src/app/pages/advisory-list/advisory-list.tsx
@@ -10,7 +10,6 @@ import {
   CardTitle,
   Grid,
   GridItem,
-  Modal,
   PageSection,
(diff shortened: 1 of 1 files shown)


gildub (approved):
LGTM

PR #128, opened 2024-07-31
Mockups for the SBOM risk profile use case

carlosthe19916:
This PR is only serving the purpose of designing and proposing changes to the UX team. Should not be merged. Creating a draft PR just for having a safe place where I can not lose the code


carlosthe19916:
closing this as it was just a POC

PR #131, opened 2024-08-07
ci: create a "publish" branch, keeping published commits

ctron:
Right now we have the issue that the tag points to a dangling commit, as the commit itself doesn't belong to a branch. Once a new tag is pushed, that commit might go away. This leads to failing local builds and failing CI release builds.

The goal of this change is to create a "publish" branch per branch that we publish, keeping the commits there.

### DIFF
--- a/.github/workflows/tag-dist.yaml
+++ b/.github/workflows/tag-dist.yaml
@@ -7,27 +7,53 @@ on:
     tags:
       - "v*"
 
+concurrency:
+  group: tag-dist
+
 jobs:
   tag-dist:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          path: build
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
           node-version: 20
       - name: Install
+        working-directory: build
         run: npm clean-install --ignore-scripts
       - name: Build
+        working-directory: build
         run: npm run build
-      - name: Commit and Tag the static dist dir
+      - name: Clone for publishing
         shell: bash
+        env:
+          SOURCE: https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}.git
         run: |
           git config --global user.name "${GITHUB_ACTOR}"
           git config --global user.email "${GITHUB_ACTOR}@users.noreply.${INPUT_ORGANIZATION_DOMAIN}"
-          git add -f client/dist
-          git commit -m "Checkin dist dir"
+          git clone "$SOURCE" publish
+          cd publish
           git fetch --tags
+          git checkout publish/${{ github.ref_name }} 2>/dev/null || git checkout -b publish/${{ github.ref_name }}
+          git merge ${{ github.ref_name }} -X theirs
+      - name: Copy dist folder
+        shell: bash
+        run: |
+          rm -Rf publish/client/dist
+          cp -a build/client/dist publish/client/
+      - name: Commit and Tag the static dist dir
+        working-directory: publish
+        shell: bash
+        run: |
+          git add -f client/dist
+          if [ -n "$(git diff --cached)" ]; then
+            git commit -m "Checkin dist dir"
+          else
+            echo "No changes detected"
+          fi
           git tag static-${{ github.ref_name }} --force
+          git push --set-upstream origin publish/${{ github.ref_name }}
           git push origin static-${{ github.ref_name }} --force
-


chirino:
LGTM.. I kinda worry about conflict merges with `git merge -X theirs static-${{ github.ref_name }}` but I guess if the ref_name stays linear we should not have issues.

chirino (approved):

carlosthe19916 (changes requested):
Wait a moment.
- This PR will generate a different branch for each commit in this repository. Am I wrong with it?
- If my previous assumption is correct, then that doesn't seem right to me. We will pollute this repository for solving a problem of an external repository. We will solve the problem of the backend creating a problem for the UI. That doesn't sound fair to me. What if the roles were switched? I honestly doubt the backend maintainers would even consider polluting their repository for solving a problem of the UI.

Proposals:
- If we are going to create a branch for every commit then those branches should be stored in a separate repository not here.
- Why do we have to do this for every commit? Just like the backend has releases, we can do releases and the backend can point to those releases (tags, not branches)

carlosthe19916:
Or maybe I am wrong...I just want to  make sure there is not a new branch for every new commit

ctron:
I need to re-work this and try it out on my forked repo first. I re-request review once I am done with that.

ctron:
> * This PR will generate a different branch for each commit in this repository. Am I wrong with it?

Yes, that's wrong. It creates one branch "ref", which is the branch name. So for each branch, we keep a "publish" branch. As we keep a single tag (per branch).

carlosthe19916:
@ctron @chirino 
What about following a more stable approach.
- The UI generates tags/releases (it could be every Thursday just like the backend).
- Before the backend do releases, the backend point to those tags/releases

It has been some time already since the backend adopted the approach of pointing to trustify-ui repo, and I saw that no matter where the trustify-ui stores its data, the backend always need to update the Cargo.lock file manually. Therefore, the backend can not catch the latest changes of the ui unless someone manually updates the Cargo.lock file. So far, I saw that the Cargo.lock is updated only when there is a backend release ongoing. Under that premise, I ask my self what is the value of generating branches in the ui every-time a commit arrives to the ui repo. Creating tags, in this repository would be enough.

My previous comment would be invalidated if we think that someone will update the Cargo.lock for every commit done in the ui. Feel free to disagree :)

ctron:
I think having a concept for releases would be great. But that would be out of scope of this PR.

What I am looking for is to have at least a stable `latest` option. Which we miss right now for creating `latest` container images.

carlosthe19916 (changes requested):
@ctron I have merge your changes to my own fork and this is what happens:
- First commit (the changes of this PR). Successfully creates  tag named: `publish/main`
- Second commit: the gh action for publishing the tag fails
- Third commit: the gh action for publishing the tag fails

You can see the last 3 commits here https://github.com/carlosthe19916/trustify-ui/commits/main/

The second and third attempts are trying to push a commit to the branch `publish/main` (it seems like that to me, but I might be wrong). If so, this is no different from the current approach but we are just changing the branch where we publish each commit from the current one to `publish/main`

carlosthe19916:
Just an idea: if we really want a "latest" thing, it should really take the latest commit from both ui and backend, otherwise we can not call it "latest" but anything else. For having a "latest", the binary/container builder should be able to take the latest commit from the ui (a similar process like the one I proposed here https://github.com/trustification/trustify/pull/639) ... If we want to create a "latest" that is going to be created by a pipeline, not a manual release, it should be fairly easy to compile the ui and backend and merge them together within the same pipeline without extra complexities like creating branches for commits in the ui repo

It was just an idea.

ctron:
Isn't that what this is trying to accomplish? Be re-using the existing build pipelines?

carlosthe19916:
It was just and idea. Feel free to accept or reject the idea.

On the other hand, please address the problem of only the first build being successfully finished as I reported in my previous comment below. Does it work in your forked branch? It doesn't in mine

> @ctron I have merge your changes to my own fork and this is what happens:
> 
> * First commit (the changes of this PR). Successfully creates  tag named: `publish/main`
> * Second commit: the gh action for publishing the tag fails
> * Third commit: the gh action for publishing the tag fails
> 
> You can see the last 3 commits here https://github.com/carlosthe19916/trustify-ui/commits/main/
> 
> The second and third attempts are trying to push a commit to the branch `publish/main` (it seems like that to me, but I might be wrong). If so, this is no different from the current approach but we are just changing the branch where we publish each commit from the current one to `publish/main`

ctron:
I am working on that.

ctron:
Ok, I think it should be fixed now.

chirino:
Yeah looks good to me.

chirino (approved):

PR #133, opened 2024-08-08
cleanup: useRouteParams now returns string or throws error

chirino:
This simplifies the queries/*.ts function signatures and implementations.

This also provides a better error messages during dev.

If you use an invalid route param arg, UI screen only displays an empty result at the current time.  Example: 

image

With this change, if you use an invalid route param, you will see:

image


carlosthe19916 (approved):
@chirino thanks for the PR. Looks good to me. Thanks for the enhancement :)

PR #134, opened 2024-08-09
ci: test image creation on prs & use Dockerfile naming

carlosthe19916:
- **ci-image-build.yaml**: If there are changes to **/package.json we test that the container images are still being able to be created. This should allow us to be more confident when upgrades of packages come.
- **Dockerfile**: Move the sentence `RUN npm install -g npm@9` from the gh action to the Dockerfile itself to be more transparent about what we are actually building.
- **Dockerfile**: Rename Containerfile by Dockerfile. I hope nobody feels offended by this. Unfortunately I found better IDE support for Dockerfile naming than for Containerfile. But I do not hold a strong opinion on this.

Also aligning the architectures of images to the ones available in the server side: Only ""amd64", "arm64""


gildub (approved):
LGTM

PR #135, opened 2024-08-09
ci: verify PRs against the e2e tests

carlosthe19916:
Based on https://github.com/trustification/trustify-ui/pull/134

- For each PR the UI is tested against the trustify-ui-tests repository (e2e tests)


chirino (approved):
It's hard test these types of PRs.   You kinda have just to commit them and see what happens and adjust as needed.   So Lgtm.

carlosthe19916:
yes, I agree. Hard to test. Thanks for approving it. If we see something wrong with it, I will take any action needed for remediating it.

PR #136, opened 2024-08-14
Use client apis more

chirino:


client/src/app/queries/sboms.ts
```diff
@@ -113,13 +116,16 @@ export const useFetchSbomsByPackageId = (
 ) => {
   const { data, isLoading, error, refetch } = useQuery({
     queryKey: [SBOMsQueryKey, "by-package", packageId, params],
-    queryFn: () => getSBOMsByPackageId(packageId, params),
+    queryFn: () =>
+      dataOf(
+        listRelatedSboms({ client, path: { id: packageId }, body: params })
```
carlosthe19916, line 121: This change is breaking the current UI.
To reproduce it:
- Go to the packages list page
- Click on any package and you will be redirected to the package details page
- Click on the `SBOMS` tab
- Open your Browser dev tools and you will see a request like:
```
GET http://localhost:3000/api/v1/sbom/by-package
```

That is wrong as the expected request should be something like:
```
GET http://localhost:3000/api/v1/sbom/by-package?limit=10&offset=0&id=85901454-4c44-5572-b3b6-7dba8aefb3e3
```

carlosthe19916 (changes requested):
@chirino please look at my comments.
Basically, currently we are losing filtering, pagination, and sorting with the current changes.

- `q` is a special query parameter that receives simple and complex query.

image

client/src/app/queries/sboms.ts
```diff
@@ -21,14 +24,14 @@ export const useFetchSBOMs = (
 ) => {
   const { data, isLoading, error, refetch } = useQuery({
     queryKey: [SBOMsQueryKey, params],
-    queryFn: () => getSBOMs(params),
+    queryFn: () => listSboms({ client }),
```
carlosthe19916, line 27: Same here, we are losing sorting, filtering, and pagination.

client/src/app/queries/sboms.ts
chirino: Thx for the heads up.  Will look into this more.

carlosthe19916: If it helps, I think we can reuse this block of code https://github.com/trustification/trustify-ui/blob/main/client/src/app/api/rest.ts#L39-L42 

That creates all the QueryParameters that axios needs/receives and appends to a request

client/src/app/hooks/table-controls/getHubRequestParams.ts
```diff
@@ -57,3 +57,35 @@ export const serializeRequestParamsForHub = (
...
+/**
+ * Like serializeRequestParamsForHub but returns a plain object instead of URLSearchParams.
+ */
+export const requestParamsQuery = (
+  deserializedParams: HubRequestParams
+): HubRequestParamsQuery => {
+  const params = serializeRequestParamsForHub(deserializedParams);
+  return {
+    limit: Number(params.get("limit")),
+    offset: Number(params.get("offset")),
```
carlosthe19916, line 87:
```
Number(undefined) = NaN
```

We should use the same logic used for `q` and `sort` where the code validates a possible undefined value of `params.get("")`

I know the code the currently calls this method always send `limit` & `offset` but better to write safe code.

client/src/app/queries/sboms.ts
```diff
@@ -113,13 +121,20 @@ export const useFetchSbomsByPackageId = (
...
     queryKey: [SBOMsQueryKey, "by-package", packageId, params],
-    queryFn: () => getSBOMsByPackageId(packageId, params),
+    queryFn: () =>
+      dataOf(
+        listRelatedSboms({
+          client,
+          path: { id: packageId },
+          query: { ...requestParamsQuery(params) },
+        })
+      ),
```
carlosthe19916, line 131: The problem reported before still exists:

> This change is breaking the current UI.
> To reproduce it:
>- Go to the packages list page
>- Click on any package and you will be redirected to the package details page
>- Click on the `SBOMS` tab
>- Open your Browser dev tools and you will see a request like:
>```
>GET http://localhost:3000/api/v1/sbom/by-package
>```
>
>That is wrong as the expected request should be something like:
>```
>GET http://localhost:3000/api/v1/sbom/by-package?limit=10&offset=0&id=85901454-4c44-5572-b3b6->7dba8aefb3e3
>```

The "packageId" should be passed as a query parameter and not as a path parameter.

carlosthe19916 (changes requested):
@chirino thanks for the last commit. I added additional comments please take a look at it when you find some time

carlosthe19916:
If it helps you can upload these files in order to see data in the "package list" => "package details" => "SBOM" tab

SBOMs:

- [quarkus-bom-3.2.11.Final-redhat-00001.json](https://github.com/user-attachments/files/16683613/quarkus-bom-3.2.11.Final-redhat-00001.json)
- [quarkus-bom-3.2.12.Final-redhat-00002.json](https://github.com/user-attachments/files/16683614/quarkus-bom-3.2.12.Final-redhat-00002.json)

Advisory:

- [CVE-2024-29025.json](https://github.com/user-attachments/files/16683616/CVE-2024-29025.json)

client/src/app/hooks/table-controls/getHubRequestParams.ts
chirino: I'll do a full implementation that avoids reusing the serializeRequestParamsForHub hack, that way it can be more efficient and more typesafe.

carlosthe19916 (approved):
LGTM! Thanks for the PR!

PR #137, opened 2024-08-21
fix: do not allow undefined values to be push to filterValues

carlosthe19916:
Make code safer and avoid potential undefined values to be set to the set of filter values set in the table-controls


chirino (approved):
Lgtm

PR #138, opened 2024-08-21
Migrate `queries/vulnerabilities.ts` to use `@app/client`

chirino:
Added a convertQuery function that converts the results of the queries to that expected type by the pages.

### DIFF
--- a/client/openapi/trustd.yaml
+++ b/client/openapi/trustd.yaml
@@ -2385,7 +2385,7 @@ components:
           sboms:
             type: array
             items:
-              $ref: '#/components/schemas/SbomStatus'
+              $ref: '#/components/schemas/VulnerabilitySbomStatus'
             description: SBOMs claimed by this advisory to be addressed by this vulnerability.
     VulnerabilityDetails:
       allOf:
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -69,5 +69,6 @@
     "ts-node": "^10.9.1",
     "type-fest": "^3.13.0",
     "typescript": "^5.5.3"
-  }
+  },
+  "packageManager": "pnpm@9.6.0+sha512.38dc6fba8dba35b39340b9700112c2fe1e12f10b17134715a4aa98ccf7bb035e76fd981cf0bb384dfa98f8d6af5481c2bef2f4266a24bfa20c34eb7147ce0b5e"
```
carlosthe19916, line 73: Just out of curiosity, what does this buy us? I've used pnpm in the past but I admit I never used it like this.

carlosthe19916 (approved):
@chirino Sorry, I needed to push the "update button" in your PR to synchronize it and make the recent "format check" script to run. 

LGTM! Feel free to push the merge button whenever you like.

chirino:
Sorry thats me using pnpm from muscle memory. I think corepack adds that so that the same pnpm version can be reused in subsequent builds. While I think pnpm lock files are one of the nicest, we likely won't be using it in production builds so I'll remove that to avoid noise.

package.json
chirino: removed /w rebase.

Issue #157, opened 2024-09-17
Create Search Page Skeleton for V1 parity


carlosthe19916:
We need to create a skeleton of the "Search page" in order to populate the search capabilities of "Sbom, Vulnerabilities (CVE), packages, and advisories".

- The page of v1 that we need to replicate can be found at https://trust.rhcloud.com/search/ 

![image](https://github.com/user-attachments/assets/1b710105-fbeb-4929-a0f1-1bda5f19818e)

- As this is just a skeleton we need to make sure of:
  - There is a single input text search where the user can type anything and when search is triggered, the value of the input text search goes down to the tab components below it (see image below please).

![Screenshot from 2024-09-17 15-39-08](https://github.com/user-attachments/assets/b9ba479e-1cb0-40cc-88bc-882e1653faef)

  - Each TAB's content is independent from the others as they have their own filters (we do not have to create any filters for the skeleton but just to keep it in mind)
  - Each TAB contains within a label the COUNT of total number of elements that matched the search criteria (the blue label next to each tab's name).

As this issue should only cover the SKELETON, anything that would allow us to have have "something" to later divide tasks by "tab" without having one tab to conflict with the development of the other tab would be enough.

PR #163, opened 2024-09-19
feat(search): add skeleton for v2 search

kahboom:
## Description
This PR introduces a skeleton page for v2 search (relates to #157). Note that this is a WIP and not feature complete.

Screenshot 2024-09-19 at 1 54 46 PM

Screenshot 2024-09-19 at 1 54 56 PM

## Changes
- Add new Search page component
- Add new route for `/search`

## TBD / To Do / In Progress
- Global search functionality for all components. Normally I might suggest using React context for global-style search, but if we're not sure whether we'll be going with dedicated pages or tabs for each searchable component, then I'll just leave it like this for now until we decide we need it. In this case, we will need to lift the state up to the common ancestral parent, `search.tsx`, and pass them down as props. @carlosthe19916 what do you think?
- Unit tests with mocks
- E2E tests


carlosthe19916:
@kahboom this PR https://github.com/trustification/trustify-ui/pull/163 was based on top of the current PR so I'm going to close this one as the other PR has been merged

Issue #161, opened 2024-09-18
UI shows "unsupported file" for compressed files


ctron:
As we do support .bz2 and .xz files, the UI should allow the user to upload such files. Currently you get:

![image](https://github.com/user-attachments/assets/5e084c5a-a046-4293-9dfb-3b1836719154)

Also see:

* https://github.com/trustification/trustify/issues/789
* https://github.com/trustification/trustify/pull/799

PR #182, opened 2024-10-04
task: Upload page redesign

### LINKED ISSUE #181: Upload V1 compatibility
Implement the Upload pages so we guarantee we can upload SBOM and Advisory files into the system.

carlosthe19916:
Fixes #181, #161

UI Tests PR: 4

- The v1 upload pages are just awful IMHO. They allow you to only upload one file at a time. Moreover, it previews the content of the file being uploaded which some times blows the browser's memory as SBOMs and Advisory files can be huge some times.

Here is a proposal for the UX team + PM to move the Upload button to the Top Nav Bar of the Page (just like Openshift does). This will allow to have visibility over the "Upload button" all the time.

> This idea was inspired by an internal chat with @chirino 

This is how it would look like 

https://github.com/user-attachments/assets/28a73de1-305f-4d06-8a35-a5b75c21380b

The current implementation has 2 tabs one for SBOMs and another Advisories.

## TODO
- If this is approved by UX then it will make sense to request the backend to just have a single REST endpoint where we can upload files so it does not matter whether or not they are an SBOM of Advisory. Users will be able to just drop their files all in a single batch
- Currently the upload endpoint does not support multipart/form which is the way of getting a  numeric progress of the upload. The current endpoint does not allow to track the progress of the upload . Unless there is a way to do it without multipart/form we will need it to be implemented in the backend.

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -17,7 +17,7 @@ ARG server_commit
 RUN dnf install git -y
 RUN git clone https://github.com/trustification/trustify.git --branch ${server_branch} /trustify/
 RUN cd /trustify/ && if [ -n "${commit}" ]; then git checkout -b commit-branch ${server_commit}; fi
-RUN sed -i 's/trustify-ui = { git = "https:\/\/github.com\/trustification\/trustify-ui.git", tag = "static-main" }/trustify-ui = { path = "..\/trustify-ui\/crate" }/g' /trustify/Cargo.toml
+RUN sed -i 's/trustify-ui = { git = "https:\/\/github.com\/trustification\/trustify-ui.git", branch = "publish\/main" }/trustify-ui = { path = "..\/trustify-ui\/crate" }/g' /trustify/Cargo.toml
 
 ######################################################################
 # Build server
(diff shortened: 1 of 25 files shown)


chirino:
Looks great!

carlosthe19916:
- @mid998 Suggested to move the "Upload" button from the top navbar to the left sidebar. I applied that change
- Allowed to support uploading .bz2 files (compressed files)
- Change the Dockerfile.server file to fix a wrong line that was not allowing the creation of the container appropriately

As part of this PR I created its e2e tests here https://github.com/trustification/trustify-ui-tests/pull/4 and this PR is using https://github.com/trustification/trustify-ui-tests/pull/4 to actually validate this PR (the Upload page).

 If in this repository we want to run the e2e tests against a PR in the trustify-ui-test repo we only need to write a comment in the description of the PR e.g.

```
UI Tests PR: 4
```

The line above means to test this PR against the PR number 4 of the trustify-ui-tests repo. Which in this case means https://github.com/trustification/trustify-ui-tests/pull/4

kahboom (approved):

Issue #171, opened 2024-10-02
CVEs List - V1 Parity


carlosthe19916:
Complete the implementation of the CVE List screen to sufficient equivalence with version 1.
Ensure that all changes in terms of fields, actions and labels have been agreed with UXD and Product Management.
Use [this sheet](https://docs.google.com/spreadsheets/d/1mhBd3-UYvSANIXwRPDrCdGplN7cq5rQPJhIywo5mcDY/edit?gid=1088076328#gid=1088076328) as a vehicle for agreeing the new collections of fields, labels and actions.

carlosthe19916:
After trying to implement this page I faced the issue of not being able to get the data required to cover some columns. The problems are described here and it needs to be discussed globally with the team [data issues](https://docs.google.com/document/d/1XNIeKHcCGWFb__Mw6Gr7h_F2u7QBBhkg86qZv6j9cC0/edit?usp=sharing)

kahboom:
completed via https://github.com/trustification/trustify-ui/pull/184

kahboom:
Reopening as some changes are needed for v1 parity

carlosthe19916:
@kahboom 

- Sorting by columns (CVSS, Date published). YES, we need that
- Missing columns
  - Related documents. THIS BECOMES SOMETHING LIKE "RELATED SBOMS"
- Date published. YES WE NEED THIS
- Remove SBOMS column. YES
- Progress bar for CVSS badge. TO HAVE CONSISTENCY OUR PM AGREED TO USE JUST THE SHIELD WE ALREADY HAVE SO NO NEED TO HAVE THE PROGRESS BAR UNLESS PM CHANGES HER MIND

PR #184, opened 2024-10-07
feat(cve): update columns

kahboom:
relates to #171 

Screenshot 2024-10-07 at 3 28 13 PM

Still need to add the Published column ~~and check the CVSS column~~.

### DIFF
--- a/client/src/app/pages/vulnerability-list/vulnerability-context.tsx
+++ b/client/src/app/pages/vulnerability-list/vulnerability-context.tsx
@@ -17,7 +17,7 @@ import { useFetchVulnerabilities } from "@app/queries/vulnerabilities";
 interface IVulnerabilitySearchContext {
   tableControls: ITableControls<
     VulnerabilitySummary,
-    "packages" | "identifier" | "title" | "severity" | "sboms" | "advisories",
+    "packages" | "identifier" | "title" | "severity" | "published" | "sboms",
     never,
     "" | "average_severity",
     string
@@ -44,12 +44,12 @@ export const VulnerabilitySearchProvider: React.FunctionComponent<
     tableName: "vulnerability",
     persistenceKeyPrefix: TablePersistenceKeyPrefixes.vulnerabilities,
     columnNames: {
-      identifier: "Identifier",
-      title: "Title",
-      severity: "Severity",
+      identifier: "Name",
+      title: "Description",
+      severity: "CVSS",
+      published: "Published",
       packages: "Packages",
       sboms: "SBOMs",
-      advisories: "Advisories",
     },
     isPaginationEnabled: true,
     isSortEnabled: true,
@@ -64,8 +64,8 @@ export const VulnerabilitySearchProvider: React.FunctionComponent<
       },
       {
         categoryKey: "average_severity",
-        title: "Severity",
-        placeholderText: "Severity",
+        title: "CVSS",
+        placeholderText: "CVSS",
         type: FilterType.multiselect,
         selectOptions: [
           { value: "none", label: "None" },
--- a/client/src/app/pages/vulnerability-list/vulnerability-table.tsx
+++ b/client/src/app/pages/vulnerability-list/vulnerability-table.tsx
@@ -94,9 +94,9 @@ export const VulnerabilityTable: React.FC = () => {
               <Th {...getThProps({ columnKey: "identifier" })} />
               <Th {...getThProps({ columnKey: "title" })} />
               <Th {...getThProps({ columnKey: "severity" })} />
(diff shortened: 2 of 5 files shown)


carlosthe19916 (commented):
- As described in the spreadsheet I think the "Advisories" Page will be removed. So rather than fixing issues there, we might take the opportunity and delete the directories:
  - /client/src/app/pages/advisory-details
  - /client/src/app/pages/advisory
And also remove the Advisory from the Routes + remove it from the sidebar. What do you think? Or is it too premature to do it? I'll leave that up to you.

I'll add more comments to the other lines in a moment :)

client/src/app/pages/vulnerability-list/vulnerability-context.tsx
```diff
@@ -88,15 +88,15 @@ export const VulnerabilitySearchProvider: React.FunctionComponent<
...
+        name: "name",
         severity: "average_score",
       },
     })
   );
 
   const tableControls = useTableControlProps({
     ...tableControlState,
-    idProperty: "identifier",
+    idProperty: "name",
```
carlosthe19916, line 99: This should remain "identifier".

- The "idProperty" points to the name of the field of an object
 E.g. Let's say the hook `useTableControlProps` is receiving a list of objects like:

```json
[
  {
     "id": 1,
     "name": "Maria",
     "surname": "Something"
  }
]
```

Then `idProperty` should point to the name of an unique field of the object being passed to the hook. Actually the IDE should be able to help there suggesting which are the available options.

In this case since we are loading Vulnerabilities then we need to pass the name of the unique (primary key) of the object, which in the case of Vulnerabilities is `identifier`

image

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
```diff
@@ -91,12 +91,12 @@ export const VulnerabilityTable: React.FC = () => {
         <Thead>
           <Tr>
             <TableHeaderContentWithControls {...tableControls}>
-              <Th {...getThProps({ columnKey: "identifier" })} />
-              <Th {...getThProps({ columnKey: "title" })} />
-              <Th {...getThProps({ columnKey: "severity" })} />
+              <Th {...getThProps({ columnKey: "name" })} />
```
carlosthe19916, line 94: The `columnKey` should be filled with the name of the "property" that was defined in the `useTableControlState` hook.

When `useTableControlState` is used we write something like the image below:

Screenshot from 2024-10-07 14-12-02

The "columnNames" represent the columns the table will have. It as a [key, value] object. The KEY represents the name of the column, and the VALUE represents how that Column will be rendered.

Example:

```
 const tableControlState = useTableControlState({   
    columnNames: {
      name: "Nombre del cliente",      
    },
```


Then we should have something like:

```
<Th {...getThProps({ columnKey: "name" })} />
```

And the Browser will render "Nombre del cliente" in the table header

carlosthe19916:
Also, the IDE is supposed to help guiding us to select the appropriate value for each field so we don't commit spelling mistakes, or wrong choices.

image

Could you confirm the IDE does not help guiding you please? Perhaps there might be something missing if that is not the case

client/src/app/pages/vulnerability-list/vulnerability-context.tsx
kahboom: Thanks @carlosthe19916, I assumed we were also changing the structure, or keys, to go from `name` to `identifier` but it seems it's just for the rendering/display, which makes sense as it seems those types are automatically generated anyway.

kahboom:
@carlosthe19916 thanks for your review, I've made the changes it builds fine now. Looking at a few things I'm missing like `Published` and the `CVSS` values showing blank, will update shortly.

>we might take the opportunity and delete the directories
>And also remove the Advisory from the Routes + remove it from the sidebar. What do you think? Or is it too premature to do it? I'll leave that up to you.

Given that we are still making a lot of changes to the design are we 100% sure we won't want to reuse that code? I don't mind either way.

>Could you confirm the IDE does not help guiding you please?

Yes, no issues with autocomplete though it's not 100% happy (see below). Not blocking me though so I'm not worrying about it for now.

Screenshot 2024-10-07 at 2 03 26 PM

carlosthe19916:
> Given that we are still making a lot of changes to the design are we 100% sure we won't want to reuse that code? I don't mind either way.

@kahboom I had a chat with our PM @mid998 and we agreed to remove the "Advisory" from the left sidebar. But let's do it in a separate PR as it will be easier to track it if at any point we need to revisit the deleted code

carlosthe19916 (approved):
LGTM! Thanks @kahboom

Issue #190, opened 2024-10-09
CVE Detail Page


kahboom:
Implement the CVE detail page

## To Do

- [x] Update overview details for CVE
- [x] `Reserved` property
- [x] Issue with tabs and background not matching
- [x] Remove search bar
- [x] CSS: fix spacing of properties in overview
- [x] Description
- [x] Update tabs for Related Products -> Related CVEs
- [x] ~~Remove Related Advisories tab~~
- [x] ~~Download button~~
- [x] Update column names for Related SBOMs
  - [x] Name
  - [x] Version
  - [x] Status
  - [x] Dependencies
  - [x] Supplier
  - [x] Created on
- [x] Update column names for Related Advisories
  - [x] Revision
  - [x] Vulnerabilities
  - [x] No aggregate severity showing up in column value?
- [x] Aesthetic enhancements (alignment of badge + title, etc.)

kahboom:
@carlosthe19916 are we getting PM to sign off on each page before closing, or can this be closed?

carlosthe19916:
@kahboom let's close it and whenever we find things to fix we can open an issue for specific things.

kahboom:
completed via https://github.com/trustification/trustify-ui/pull/211

PR #192, opened 2024-10-10
feat(cve): details page v1 parity

kahboom:
implements #190 

## Changes

- Rearrange properties in `Overview` component to reflect v1
- Adapt labels in tables according to [this guide](https://docs.google.com/spreadsheets/d/1mhBd3-UYvSANIXwRPDrCdGplN7cq5rQPJhIywo5mcDY/edit?gid=1088076328#gid=1088076328)
- Move tabs + `Overview` into dedicated area outside of `DetailPage`
- Remove `DetailPage`, as we are no longer using those breadcrumbs, built-in tabs, or actions

## Comparison of v1 and v2

v1v2

## New state

Advisories

SBOMs

Packages

Detail popover

## TBD

- Keep or ditch the "Related Products" tab from v1
- Keep or ditch "Packages" tab from v2
- Do we want to show properties not shown (available?) in v1 ("non-normative", "CWE", "withdrawn", "released")?
- What happened to the "Reserved" property in v1, was that renamed? maybe to one of the above?

### DIFF
--- a/client/src/app/pages/vulnerability-details/overview.tsx
+++ b/client/src/app/pages/vulnerability-details/overview.tsx
@@ -1,17 +1,12 @@
 import React from "react";
 
 import {
-  Card,
-  CardBody,
-  CardTitle,
   DescriptionList,
   DescriptionListDescription,
   DescriptionListGroup,
   DescriptionListTerm,
-  Grid,
-  GridItem,
-  Stack,
-  StackItem,
+  Text,
+  TextContent,
 } from "@patternfly/react-core";
 
 import { formatDate } from "@app/utils/utils";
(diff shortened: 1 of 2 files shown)


client/src/app/pages/vulnerability-details/overview.tsx
```diff
@@ -24,76 +19,52 @@ interface OverviewProps {
...
+        <DescriptionListGroup>
+          <DescriptionListTerm>Withdrawn</DescriptionListTerm>
+          <DescriptionListDescription>
+            {formatDate(vulnerability.withdrawn)}
+          </DescriptionListDescription>
+        </DescriptionListGroup>
+        <DescriptionListGroup>
+          <DescriptionListTerm>CWE</DescriptionListTerm>
+          <DescriptionListDescription>
+            {"vulnerability.cwe"}
```
carlosthe19916, line 58: I think the property is `cwes`. And this is a mistake in the current main branch as it is rendering a string and not the value of the variable

carlosthe19916 (commented):
Thanks for the PR @kahboom here are some minor points I noticed:
- The new Vulnerability Details page does not have the Breadcrumb to navigate to the previous page as the other pages has. 
- The new Tabs have a grey background and not white as other pages. I am not saying the white one is right or wrong but perhaps we need to make an agreement on how all details pages should look like (either white or gray + any other common patterns) so all pages are consistent. We could do that in another PR surely so we don't block this PR. 
- Because of the gray background there is also a grey space between the table and the bottom pagination. That doesn't look nice IMHO.

Screenshot from 2024-10-11 08-59-28

And to your questions:

> Do we want to show properties not shown (available?) in v1 ("non-normative", "CWE", "withdrawn", "released")?

I don't know. I would suggest to keep that information. Show it to UX and PM, take a decision and move on.

> What happened to the "Reserved" property in v1, was that renamed? maybe to one of the above?

Good question. Unfortunately, I do not know the answer. We need to ask this question to the backend guys as this is a technical thing. Perhaps the "reserved" property is now either "Modified" or "Withdrawn"

kahboom:
@carlosthe19916 sorry for the confusion, I thought we were also wanting it to be visually the same as it was before? like [this](https://trust.rhcloud.com/cve/content/CVE-2023-26045):

Screenshot 2024-10-11 at 11 28 03 AM

That's why I removed the breadcrumbs and the `DetailPage` etc. but I can revert that. What you showed is the SBOM "related vulnerabilities", so we should be using that same style/layout for CVEs/vulnerabilities detail as well?

carlosthe19916 (approved):
Thanks @kahboom if we find further changes in these pages we can open new PRs. Overall look good, thanks!

PR #195, opened 2024-10-11
fix: rever http-proxy-middleware version and make sure the container works with it

carlosthe19916:
The PR https://github.com/trustification/trustify-ui/pull/99 upgraded http-proxy-middleware to version 3 as a consecuence the container image does not work anylonger.

How to see the error:
- Start the backend with the instructions from https://github.com/trustification/trustify 

```shell
cargo run --bin trustd
```

- The following command was extracted from the backend README:
```shell
podman run --network="host" --pull=always \
-e TRUSTIFY_API_URL=http://localhost:8080 \
-e OIDC_CLIENT_ID=frontend \
-e OIDC_SERVER_URL=http://localhost:8090/realms/trustify \
-e ANALYTICS_ENABLED=false \
-e PORT=3000 \
-p 3000:3000 \
ghcr.io/trustification/trustify-ui:latest
```

- Open the UI at http://localhost:3000

An error will appear in the ui container and the UI can not connect to the backend.

This PR is reverting the version upgraded and makes sure the container image goes back to work.
We could upgrade our dependencies but without breaking what works already

### DIFF
--- a/client/src/app/hooks/table-controls/filtering/useFilterState.ts
+++ b/client/src/app/hooks/table-controls/filtering/useFilterState.ts
@@ -72,7 +72,7 @@ export const useFilterState = <
 
   if (isInitialLoad) {
     initialFilterValues = isFilterEnabled
-      ? args?.initialFilterValues ?? {}
+      ? (args?.initialFilterValues ?? {})
       : {};
   }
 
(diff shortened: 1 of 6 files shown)


kahboom (approved):
looks good to me @carlosthe19916 , apologies for the delay as it took me a bit to reproduce and verify. thanks!

PR #218, opened 2024-11-05

carlosthe19916:

### DIFF
--- a/client/openapi/trustd.yaml
+++ b/client/openapi/trustd.yaml
@@ -3630,6 +3630,7 @@ components:
       - identifier
       - title
       - description
+      - reserved
       - published
       - modified
       - withdrawn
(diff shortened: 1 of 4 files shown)


carlosthe19916:
@kahboom this is an automated PR generated by the backend after the openapi.yaml file changes.
We use the openapi.yaml file to generate the REST client in this repository.

I think you requested the "reserved" field to be added to vulnerabilities therefore the openapi definition has changed. However, there is an error in one of the stories. 

If you need to get the "reserved" field in the UI you can merge this PR but we will need to fix the ci build errors first

carlosthe19916:
IMO although painful it is good to have errors also in storybook. If we do not have errors we won't be able to fix them as those errors might be overlooked.

- There will be PRs like this one https://github.com/trustification/trustify-ui/pull/210 that touch the story files. And even the "preview" might not work and the PR creator might be wonder why
- We also publish stories at http://www.trustification.io/trustify-ui/ and we could also wonder why that page is not updated; later we might discover that it was due to a warning we might've ignored.

I would say better to be safe even if it takes us more time to maintain the code

kahboom (approved):

PR #229, opened 2024-11-08
fix(cve): CVE list fixes & enhancements

kahboom:
Relates to #171 

## Changes

- Add back in missing properties
- Add CVE list components

### DIFF
--- a/client/src/app/pages/vulnerability-list/vulnerability-context.tsx
+++ b/client/src/app/pages/vulnerability-list/vulnerability-context.tsx
@@ -18,7 +18,7 @@ export interface IVulnerabilitySearchContext {
   tableControls: ITableControls<
     VulnerabilitySummary,
     "identifier" | "title" | "severity" | "published" | "sboms",
-    never,
+    "published" | "sboms",
     "" | "average_severity",
     string
   >;
@@ -52,7 +52,7 @@ export const VulnerabilitySearchProvider: React.FunctionComponent<
     },
     isPaginationEnabled: true,
     isSortEnabled: true,
-    sortableColumns: [],
+    sortableColumns: ["published", "sboms"],
     isFilterEnabled: true,
     filterCategories: [
       {
@@ -87,8 +87,8 @@ export const VulnerabilitySearchProvider: React.FunctionComponent<
     getHubRequestParams({
       ...tableControlState,
       hubSortFieldKeys: {
-        identifier: "identifier",
-        severity: "average_score",
+        published: "published",
+        sboms: "sboms",
       },
     })
   );


kahboom:
@carlosthe19916 question for Monday -- any idea why the mocked context isn't working for the list view? do I need to mock the entire table library? I'm trying to follow this guide but without much luck: https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-providers

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
```diff
@@ -32,6 +32,7 @@ export const VulnerabilityTable: React.FC = () => {
       getTdProps,
     },
   } = tableControls;
+  console.table(tableControls);
```
carlosthe19916, line 35: We could remove this

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
```diff
@@ -54,6 +55,7 @@ export const VulnerabilityTable: React.FC = () => {
           numRenderedColumns={numRenderedColumns}
         >
           {currentPageItems.map((item, rowIndex) => {
+            console.table(currentPageItems);
```
carlosthe19916, line 58: we could remove this too

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
```diff
@@ -74,11 +76,7 @@ export const VulnerabilityTable: React.FC = () => {
                         {item.identifier}
                       </NavLink>
                     </Td>
-                    <Td
-                      width={45}
-                      modifier="truncate"
-                      {...getTdProps({ columnKey: "title" })}
-                    >
+                    <Td width={45} {...getTdProps({ columnKey: "title" })}>
```
carlosthe19916, line 79: The `modifier=truncate` has been removed. Any particular reason? Initially "truncate" was set expecting to have long strings there and render a tooltip.

Ahh it might be that V1 didn't have that column truncated? But as far as I remember, in v1 we didn't apply truncate not because we didn't want to but because we couldn't do it so it was a limitation and not what we desired to have.

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
```diff
@@ -97,7 +95,7 @@ export const VulnerabilityTable: React.FC = () => {
                         rowIndex,
                       })}
                     >
-                      X
+                      {item.cwes.length}
```
carlosthe19916, line 98: I think we should render not the length but the content of the array. If there are multiple elements we could join them using comma.

carlosthe19916 (changes requested):
@kahboom Thanks for the PR. Some minor things to change in this PR.

Besides that, to your question about mocking the table library: I do not think that is the way to go; that would be too much. I'll find some time this week to look into it and find out where the issue is :)

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
kahboom: @carlosthe19916 yep the reasoning was because it wasn't there in v1, but also sometimes large chunks of text having a big tooltip felt awkward, but I will put it back because I agree with you, and think is does provide more benefit than any inconvenience

kahboom:
Updated with sorting enabled:

Screenshot 2024-11-11 at 3 42 13 PM

carlosthe19916 (approved):

carlosthe19916:
Sorry for the delayed review

PR #238, opened 2024-11-13
fix: CVE Details Page - SBOM/Advisory should redirect to pages

carlosthe19916:
CVE Details page:

- Related SBOMs Tab

The Sbom Name table column should redirect the user to the SBOM Details page. I replaced the current button that opens a Drawer by a Link that should redirect the user. This should be consistent with V1 parity

- Related Advisories Tab

The current ADvisory ID table column redirects the user to `/advisory/{id}` but it should be `/advisories/{id}` as that is the naming we are using in our Routes

Also taking the change to delete `client/src/app/components/SbomInDrawerInfo.tsx` as we don't need it now and we don't know if we ever will need it. Better to keep the code clean


kahboom (approved):
thanks @carlosthe19916! sorry I missed those bits

Issue #235, opened 2024-11-12
Vulnerabilities List Page: Related documents column


carlosthe19916:
Currently we are rendering an "X" into that column.

We need to render the number of SBOMS affected by the Vulnerability

kahboom:
This should be done already through #229 , however, I am not able to get sample data to show a meaningful value in the UI:

![Screenshot 2024-11-13 at 11 02 43 AM](https://github.com/user-attachments/assets/53e9618d-1c88-4c67-9ee8-efed9ca6ff85)

![Screenshot 2024-11-13 at 11 06 00 AM](https://github.com/user-attachments/assets/343450c1-65ce-455b-85f0-db7c8981e41b)

And when I try to sort by "Related documents" I'm getting an error in the browser console about the column name being invalid, so I'll have to check that out:

![Screenshot 2024-11-13 at 11 08 33 AM](https://github.com/user-attachments/assets/c562efaa-3a52-4e27-940c-c3cfc613d134)

carlosthe19916:
@kahboom I missed https://github.com/trustification/trustify-ui/pull/229 sorry.

But:

- Since in V1 that column is not sortable perhaps we can just remove the sorting feature of that column to avoid issues
- The "Related Documents" columns is quite ambiguous in V1.
  - Let's ask PM to verify what exactly she expects to see in that column. The current code is rendering the count of CWEs but I am thinking that perhaps it should render the number of SBOMs affected by that Vulnerability and not CWEs. We'll need to confirm it

PR #239, opened 2024-11-13
fix(cve): change sorting columns in list view

kahboom:
Just a quick fix related to #235. This PR updates the sortable columns in CVE list view to be `published` and `severity`, and removes sorting by `Related documents`. Note that the default sort has been set to `published` because it seems to not be possible to sort by `severity`. Will check to see if this is supported by the backend.

Screenshot 2024-11-13 at 11 54 47 AM

### DIFF
--- a/client/src/app/pages/vulnerability-list/vulnerability-context.tsx
+++ b/client/src/app/pages/vulnerability-list/vulnerability-context.tsx
@@ -18,7 +18,7 @@ export interface IVulnerabilitySearchContext {
   tableControls: ITableControls<
     VulnerabilitySummary,
     "identifier" | "title" | "severity" | "published" | "sboms",
-    "published" | "sboms",
+    "published" | "severity",
     "" | "average_severity",
     string
   >;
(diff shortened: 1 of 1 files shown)


client/src/app/pages/vulnerability-list/vulnerability-context.tsx
```diff
@@ -87,8 +87,8 @@ export const VulnerabilitySearchProvider: React.FunctionComponent<
     getHubRequestParams({
       ...tableControlState,
       hubSortFieldKeys: {
+        severity: "severity",
```
carlosthe19916, line 90: To sort by severity we could change this line to `severity: "average_severity"`.

The `hubSortFieldKeys` contains a key value object where the KEY is always the fields we defined in the UI. And the VALUE is the backend filter key we want to use.

carlosthe19916 (changes requested):

client/src/app/pages/vulnerability-list/vulnerability-context.tsx
carlosthe19916: `severity: "average_score"` should also be equivalent to `severity: "average_severity"`

kahboom: ah I see, I'll get that updated now

kahboom:
@carlosthe19916 thanks, sorting works now since updating it to use `average_severity`:

Screenshot 2024-11-13 at 2 12 47 PM

carlosthe19916 (approved):

Issue #234, opened 2024-11-12
Vulnerabilities List page: Filter by "Published"


carlosthe19916:
The V1 page has a filter applied to the "Published" table column.

We need to implement a filter for it. But we agreed with PM that date filters should allow the user to select a range and not predefined values. Actually the SBOM List page already has that filter so it might be a matter of copying and pasting it.

![image](https://github.com/user-attachments/assets/57e1135c-9234-4b5c-add2-f1b731df597d)

carlosthe19916:
Fixed by https://github.com/trustification/trustify-ui/pull/242

PR #242, opened 2024-11-14
feat(cve): add filter by published date

kahboom:
Implements #234 

- Add filter by CVE `published` property
- Update tab in Search page to use `CVEs` instead of `Vulnerabilities`

Screenshot 2024-11-14 at 10 10 09 AM

Screenshot 2024-11-14 at 10 13 03 AM


carlosthe19916 (approved):
LGTM! Thank you very much @kahboom

PR #252, opened 2024-11-18
feat: persist search table filters in the browser urls

carlosthe19916:
Fixes: https://github.com/trustification/trustify-ui/issues/198

The filters are now saved in the browser url allowing it to be restored when the user click on the Browser's back button.

https://github.com/user-attachments/assets/646e32da-9837-4afd-b7d7-99bc2763704f


kahboom (approved):
awesome!

PR #254, opened 2024-11-19
fix: render "Rejected" prefix if Vulnerability has "withdrawn" date

carlosthe19916:
Fixes: https://github.com/trustification/trustify-ui/issues/90


kahboom (approved):
good call moving this to its own new component, it will make it easier for stories + testing going forward as well

PR #266, opened 2024-11-22
fix: sbom vulnerabilities correct grouping

carlosthe19916:


carlosthe19916:
Fixes: https://github.com/trustification/trustify/issues/1046

PR #267, opened 2024-11-22
fix: upload progress does not have blue color. Force it

carlosthe19916:
fixes: https://github.com/trustification/trustify-ui/issues/265

Currently Patternfly React does not allow to use the BLUE color for the "In progress" bars. I reported this bug in their repository https://github.com/patternfly/patternfly-react/issues/11276 but that will take time to be fixed.

@jcrossley3 the main reason of the issue was that although a file cam be 100% sent to the server, the server can still take time to process the file although the file has been fully sent already. This PR should make the Progress Bar always BLUE unless the server gives a response. It also adds a helper text to explain that.

https://github.com/user-attachments/assets/d5a02a23-ee0c-41d1-98a3-c0205473bb9b

### DIFF
--- a/client/src/app/App.css
+++ b/client/src/app/App.css
@@ -12,3 +12,16 @@
 table.vertical-middle-aligned-table tr td {
   vertical-align: middle;
 }
+
+.multiple-file-upload-status-item-force-blue .pf-v5-c-progress.pf-m-success {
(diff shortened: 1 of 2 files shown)


jcrossley3 (approved):
I didn't try it locally, but if it works like the video, that's perfect, thanks!

Issue #273, opened 2024-11-25
Implement the "Input Search" with  "Dropdown" that we have in V1


carlosthe19916:
- The following image is the Search Page
- On the top right side there is an input text that renders a dropdown while the user types any string on it. It call SBOM + Package + Vulnerability + Advisory endpoints
- The user can click on any of the elements of the Dropdown. If the user clicks on an element of the Dropdown then the user is redirected to the Details page of the element selected.

![image](https://github.com/user-attachments/assets/1ed6dd55-ec3a-430c-aac1-5d98d7762988)

carlosthe19916:
@kahboom @gildub may I ask your help with this? First arrives, first served (assign yourself the issue)

I need to jump into a task in V1 (Download License). If this issue is not taken, I can take it myself after I finish the V1 task :)

kahboom:
Sorry @carlosthe19916 I've literally just seen this. I'm working on the conveyor belt and story for products but this looks like it takes priority so I can take a look tomorrow.

kahboom:
@carlosthe19916 are we fine to close this or do we need any more changes?

carlosthe19916:
@kahboom yes, this should be close.
I was expecting that once the PR linked to this issue is merged then this issue could be automatically closed. So let's close it

PR #277, opened 2024-11-28
feat(search): add input search

kahboom:
This PR implements #273 by adding the input search dropdown/typeahead menu for the Search Page

## Changes

- Add new hook to hold state of pre-formatted entities rendered in dropdown
- Add new `SearchMenu` component with local state + PF search input

## Screenshots

Screenshot 2024-11-28 at 1 46 54 PM

Screenshot 2024-11-28 at 1 51 23 PM

Screenshot 2024-11-28 at 1 45 11 PM


client/src/app/pages/search/components/SearchMenu.tsx
```diff
@@ -0,0 +1,314 @@
...
+    | "gold"
+    | undefined;
+}
+
+function useAllEntities() {
+  const [entityList, setEntityList] = React.useState<IEntity[]>([]);
+
+  const {
+    tableControls: { currentPageItems: advisories },
+  } = React.useContext(AdvisorySearchContext);
```
carlosthe19916, line 40: This is extracting the list of advisories of the SearchPageContext but I think we should change it:

- Let's say the Backend has one thousand `Advisories`
- The `AdvisorySearchContext` contains the information of the data to be rendered in the table (10 elements)
- If extract data from `AdvisorySearchContext` here then it means the dropdown will only contain elements from the current table. But I think what should be done is to search in the backend according to what the user types regarless of what the values of `AdvisorySearchContext` are.

So this block of code could be replaced by:

```typescript
function useAllEntities(filterText: string) {
  const [entityList, setEntityList] = React.useState<IEntity[]>([]);

  const {
    result: { data: advisories },
  } = useFetchAdvisories({
    filters: [
      { field: FILTER_TEXT_CATEGORY_KEY, operator: "~", value: filterText },
    ],
    page: { pageNumber: 1, itemsPerPage: 10 },
  });
```

In the code I am sharing the function `useAllEntities` receives a param `filterText` which represents what the user types in the input text, then that string is send to the backend to do a filter among the hypothetical thousand Advisories in the database

Same logic can be applied to the other 3 entities of this function (Packages, Sboms, Vulnerabilities)

client/src/app/pages/search/components/SearchMenu.tsx
```diff
@@ -0,0 +1,314 @@
...
+    } else {
+      setIsAutocompleteOpen(false);
+    }
+
+    setSearchValue(newValue);
+  };
+
+  const onClearSearchValue = () => {
+    setSearchValue("");
+    onChangeSearch("");
```
carlosthe19916, line 194: Perhaps `onChangeSearch("")` should not be called here? I tried to test it and when I start typing text and then click on the "X" icon that appears on the right of the input text then a filter "''" (empty string) is applied. But we should wait for the user to press down "Enter" or click on the "Search" button before applying a filter.

client/src/app/pages/search/components/SearchMenu.tsx
```diff
@@ -0,0 +1,314 @@
...
+    onChangeSearch(searchValue);
+  };
+
+  // Whenever an autocomplete option is selected, set the search input value, close the menu, and put the browser
+  // focus back on the search input
+  const onSelect = (e: any, itemId?: string | number | undefined) => {
+    e?.stopPropagation();
+    setSearchValue(itemId as string);
+    setIsAutocompleteOpen(false);
+    searchInputRef.current?.focus();
```
carlosthe19916, line 207: It might be just my eyes but I feel like there is a slight delay between the action CLICK and the redirect. Perhaps we could directly call `navigate("/new/url")` here? but I didn't test whether or not that would make it faster.

https://github.com/user-attachments/assets/984fb91b-0ec9-409c-8e53-633d394817ae

client/src/app/pages/search/components/SearchMenu.tsx
```diff
@@ -0,0 +1,314 @@
...
+          ...entityList
+            .filter(
+              (option: IEntity) =>
+                !option.id.startsWith(newValue.toLowerCase()) &&
+                option.id.includes(newValue.toLowerCase())
+            )
+            .map((option: IEntity) => (
+              <MenuItem itemId={option.id} key={option.id}>
+                {option.id}
+              </MenuItem>
```
carlosthe19916, line 177: The <MenuItem> differs from the <MenuAction> on line 153 therefore how it is rendered is completely different as its predecessor. See in the video below that there are elements in the dropdown that only have an ID rendered in the Dropdown and not label. As a use it might be difficult to see IDs rendered in the Dropdown

https://github.com/user-attachments/assets/8a94e655-562e-4555-bdbe-91f9b0796724

client/src/app/pages/search/components/SearchMenu.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+      // When the value of the search input changes, build a list of no more than 10 autocomplete options.
+      // Options which start with the search input value are listed first, followed by options which contain
+      // the search input value.
+      let options: React.JSX.Element[] = entityList
+        .filter(
+          (option) =>
+            option.id.toLowerCase().startsWith(newValue.toLowerCase()) ||
+            option.title?.toLowerCase().startsWith(newValue.toLowerCase()) ||
+            option.description?.toLowerCase().startsWith(newValue.toLowerCase())
```
carlosthe19916, line 150: I don't have anything in mind just yet but I wonder if the filtering logic could be done with a "contains" and not "startWith". Maybe in a second PR we can come up with some logic to appropriately filter by string. but we can leave this for later I think

carlosthe19916 (changes requested):
@kahboom thank you for the PR! Looks great
I added some observations I made while testing your PR. Please take a look at them

client/src/app/pages/search/components/SearchMenu.tsx
kahboom: Yes I see it as well, and I did notice at the time but forgot to go back to it! I'll try it with `navigate`

client/src/app/pages/search/components/SearchMenu.tsx
kahboom: I moved the filter functionality into its own function, and I created a test for it. Then I created a story that shows how you can override the filter function, so we can show what it would look like if we change the logic. And there is an example with using `includes` instead of `startWith` but I can come back to it and make it better later on.

client/src/app/pages/search/components/SearchMenu.tsx
kahboom: I couldn't fix this, even with `useNavigate`, so I'll be creating an issue to track it

client/src/app/pages/search/components/SearchMenu.tsx
```diff
@@ -0,0 +1,329 @@
...
+        option.id.toLowerCase().startsWith(searchString.toLowerCase()) ||
+        option.title?.toLowerCase().startsWith(searchString.toLowerCase()) ||
+        option.description?.toLowerCase().startsWith(searchString.toLowerCase())
+    )
+    .map((option) => (
+      <MenuItem
+        itemId={option.id}
+        key={option.id}
+        description={option.description}
+        to={option.navLink}
```
carlosthe19916, line 53: I discovered the reason of the delay while redirecting to pages. the `to` element is causing a "Full page reload". This should be replaced by the native react `useNavigate` or a `<Link/>` component.

client/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
-  "include": ["src/**/*", "config/**/*", "types/**/*"],
+  "include": ["src/**/*", "config/**/*", "types/**/*", "setupTests.ts"],
```
carlosthe19916, line 3: I tested your PR removing this change and everything worked. Is there a reason to add this change? I mean, all tests seem to be executed currently without this change

carlosthe19916 (changes requested):
@kahboom thanks for the enhancements.

When I type "a" i see some Elements like "item.decomposedPurl?.namespace" being printed in the Dropdown component (see image below). I think that is a mistake. I know packages are tricky as even V1 do not include them in the dropdown. I don't think it would be wrong also to remove it if you consider it is better.

Screenshot From 2024-12-02 10-27-24

client/tsconfig.json
kahboom: This was a mistake, sorry I didn't catch it in my self review, going to remove it now. Thanks!

client/src/app/pages/search/components/SearchMenu.tsx
kahboom: For me it still shows a delay with either one (a bit improved though), but I will add it and we can always open an issue if you are still seeing it happening.

kahboom:
@carlosthe19916 thanks for the review!

>When I type "a" i see some Elements like "item.decomposedPurl?.namespace" being printed

Oops, this is how it should look now:

Screenshot 2024-12-02 at 11 03 43 AM

>I don't think it would be wrong also to remove it if you consider it is better.

I've decided to keep them because we decided with Michelle to just show everything and we narrow stuff later on. Let me know how you feel about the changes.

carlosthe19916 (commented):
@kahboom I created a PR to your PR here https://github.com/kahboom/trustify-ui/pull/1

It should address the delay thing. To me, that is as fast as any other redirect we might have in other parts of the page.

Please take a look at it

carlosthe19916 (approved):

PR #298, opened 2024-12-13
feat: add Vuln Details -> Related SBOMBs -> dependencies list

carlosthe19916:
Matching changes done in the backend at https://github.com/trustification/trustify/pull/1098

Now we should be able to render the packages within an SBOM when we are in the VUlnerabilities page Details. See image below... it should also be consistent with V1 as V1 has exactly the same expanded area on the "Dependencies column"

image


dejanb (approved):
Works like a charm

PR #326, opened 2025-01-29
fix: sbom's packages view

carlosthe19916:
In response to https://github.com/trustification/trustify/issues/1163

The UI now should use the QueryParam `purl` to get the sboms related to a package

image


gildub (approved):
LGTM

PR #328, opened 2025-01-29
Add support to latest NodeJS LTS version

gildub:
NodeJS 22 is LTS

- Using UBI9/nodejs-22 for Dockerfiles
- Updated github workflows accordingly
- Bump @hey-api clients to support nodejs 22
- Bump few @storybook packages for proper support of nodejs 22 types
 
Fix https://github.com/trustification/trustify-ui/issues/327


carlosthe19916 (changes requested):
The test are failing. And it is due to the changes in this PR. See this other test PR I created https://github.com/trustification/trustify-ui/pull/330 , the tests are passing but they fail here. https://github.com/trustification/trustify-ui-tests has quite a simple test so there is something wrong with this PR. Please clone that repo and make sure this PR does not break that basic test... on a side note, I am happy that we have the e2e tests even at a basic level, we can capture these kind of things, that hopefully those test will get even better.

On the other hand. If I execute:
```shell
npm run start:dev
```

In a clean environment I get the error:
```
Cannot find module '@app/client' or its corresponding type declarations
```

That is caused because we are not including `client/src/app/client/` anymore. While I see the reason behind ignoring that folder, seems like it is not actually solving a problem but creating one.

carlosthe19916:
@gildub here is the PR I mentioned internally before https://github.com/trustification/trustify-ui/pull/372 
As mentioned the only dependency that required to be upgraded was @storybook/addon-webpack5-compiler-swc 

Feel free to close my PR and take it just as an example. I don't want to step of your toes, just to share what I have discovered while playing with it.

carlosthe19916 (changes requested):
@gildub This is my environment

```
NodeJS v22.14.0
NPM v10.9.2
```

I checkout this PR. Then 

```sh
npm ci --ignore-scripts
```

Then
```sh
npm run start:dev
```


https://github.com/user-attachments/assets/70154601-028e-4d17-aeab-ae7b55bb0432

- The dev mode is broken
- Clearly the ci pipeline fails for a reason

gildub:
@carlosthe19916, this is ready to go, please confirm.

carlosthe19916:
> @carlosthe19916, this is ready to go, please confirm.

Give me some time please.

carlosthe19916 (approved):
LGTM!
We need to synchronize this important change here https://gitlab.cee.redhat.com/trustification/rhtpa-konflux-release.0.2.z/-/blob/main/Dockerfile.server?ref_type=heads#L4

@gildub could you take care of it please?
- First the downstream image needs to be changed.
- Only after it this PR can be merged otherwise we will destroy the downstream pipeline

PR #331, opened 2025-01-31
Avoid saving files generated by trustd.yaml | "openapi spec" into git

gildub:
Files that are generated, for instance by @hey-api, don't have to be committed to the repo meanwhile we have to make sure they're added by the build.

This also fix the Know issue described in the README.md file about missing module `@trustify-ui/common`.

On the backend, there is no need to trigger new file generation when changes are applied to openAPI : https://github.com/trustification/trustify/pull/1242


.github/workflows/ci-global.yaml
```diff
@@ -18,6 +18,18 @@ jobs:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+  
+      - name: Install dependencies
+        run: npm install
+
+      - name: Run build
+        run: npm run build 
+
```
carlosthe19916, line 32: This is not needed

package.json
```diff
@@ -10,12 +10,11 @@
...
     "format:check": "npx prettier --check './**/*.{ts,tsx,js,json}'",
     "format:fix": "npx prettier --write './**/*.{ts,tsx,js,json}'",
-    "openapi:generate-client": "npm run openapi:generate-client -w client",
-    "generate": "npm run openapi:generate-client",
+    "generate": "npm run generate -ws --if-present",
     "clean": "rimraf ./dist && npm run clean -ws --if-present",
     "clean:all": "npm run clean && rimraf ./node_modules ./**/node_modules/",
     "dist": "rimraf ./dist && copyfiles -e 'node_modules/**' entrypoint.sh '**/package.json' '*/dist/**/*' ./dist",
-    "build": "npm run build -ws --if-present",
+    "build": "npm run generate && npm run build -ws --if-present",
```
carlosthe19916, line 17: With the CURRENT changes, this is not needed technically because Your PR is not ignoring the files from `hey-api` then this is not needed.

carlosthe19916 (changes requested):
2 points:
- From the internal chat my impression was that you want this PR for ignoring the `hey-api` files. If that is your intention I think you should ignore them in this PR and also open a PR in the backend to stop sending the generated files to this Repo but only send the `trustd.yaml` file. The order should be: First the backend PR is merged and only then this PR can be merged.

- The second point: once you deleted the `hey-api` generated files make sure ALL commands work not just the build one. Then test yourself the experience. E.g. if we generate the `hey-api` files each time `npm run start:dev` is executed then feel the experience, to me feel like it slows me down rather than helping me but I'd like to know your experience

.github/workflows/ci-global.yaml
gildub: This will be needed if we remove the generated files.

gildub: @carlosthe19916, 

Point 1 : I agree with you, it makes sense to ignore generated files (form @hey-api) in this PR.

Point 2 : The files must be generated during the build not at start.

package.json
gildub: That's right. But will be needed with next commit.

.github/workflows/ci-global.yaml
carlosthe19916: Just to make sure I am conveying the right message. I think this is not needed because the code below your changes are:

```yaml
      - name: save trustify-ui image
        run: |
          docker build . -t ghcr.io/trustification/trustify-ui:pr-test -f Dockerfile
          docker save -o /tmp/trustify-ui.tar ghcr.io/trustification/trustify-ui:pr-test
```

Which is building a container image based on a Dockerfile.
A Dockerfile MUST BE SELF SUFFICIENT. a Dockerfile should not relay on previous commands being executed. E.g.

- Clone the repo (clean copy). Do not execute any command. As a matter of fact you don't even need Node to be installed in your machine then:
- `docker build . -t my_image -f Dockerfile` MUST work. If we break that basic principle then we are not doing the right change.

What this PR Changes is telling me is that Before I build the container I need to execute `npm run build` first, which is unacceptable.

gildub: You're right I misinterpreted the way the image is build here. Let me revisit that....

carlosthe19916:
- @gildub have you tried `npm run start:dev`? or `npm run storybook`? Please make sure ALL commands work as expected not just the `build` command.
- Please I'd like you to have the experience of executing `npm run start:dev` multiple times as any average developer would do during the day. What is your experience? Do you think generating the `hey-api` slows you down? or it does not matter at all? are there possible ways of generating the `hey-api` files only when strictly needed?

If you do the 2 steps above and we are happy with it, open a PR in the backend to change https://github.com/trustification/trustify/blob/main/.github/workflows/openapi.yaml#L27-L35 accordingly.

carlosthe19916 (approved):
@gildub looks good to me.
Please, do not merge this PR until we also have an approved or merged PR in the backend, to be specific align these changes at https://github.com/trustification/trustify/blob/main/.github/workflows/openapi.yaml. Otherwise we are going to keep receiving this kind of PRs https://github.com/trustification/trustify-ui/pull/337

PR #332, opened 2025-02-03

carlosthe19916:


gildub:
@carlosthe19916, this PR won't be needed once PR#331 has been merged.
As a matter of fact there will be no need to update files generated by @hey-api.

carlosthe19916:
@gildub this is an auto generated PR coming from the backend repository.
I think we are safe merging this because anyway #331 will delete those files but in the meantime we still rely on these files.

PR #339, opened 2025-02-05
Update tag-dist.yaml

carlosthe19916:
Move - "release/*" to branches and not to releases


mrizzi (approved):
The second is better 😉

PR #341, opened 2025-02-05
ci: add backport action

carlosthe19916:
solves https://github.com/trustification/trustify-ui/issues/336

Any PR that needs to be backported to the branch release/0.2.z need to add the label "backport release/0.2.z" manually, then a backport PR should be created by gh actions.

I tested it in my own account see:
- https://github.com/carlosthe19916/trustify-ui/pull/22 is a PR opened from a forked repository
- a backport https://github.com/carlosthe19916/trustify-ui/pull/23 was created automatically

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,29 @@
+name: Backport
+on:
+  pull_request_target:
+    types: [ closed ]
+  issue_comment:
+    types: [ created ]
+permissions:
+  contents: write
+  pull-requests: write
+jobs:
+  backport:
+    name: Backport pull request
+    runs-on: ubuntu-latest
+    if: >
+      (
+        github.event_name == 'pull_request_target' &&
+        github.event.pull_request.merged
+      ) || (
+        github.event_name == 'issue_comment' &&
+        github.event.issue.pull_request &&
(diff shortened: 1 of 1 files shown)


carlosthe19916:
The downside of this approach is that Github Actions on the backported PR won't run because a PR opened by github itself cannot trigger the actions. The solution to that is that each time the backport PR is opened, the reviewer Opens and closes the PR again to trigger the actions. See https://github.com/orgs/community/discussions/55906

gildub (approved):
Having to manually intervene defeats the purpose of the automation but that's much faster than to create the backport itself and it will eventually go away.

carlosthe19916:
> Having to manually intervene defeats the purpose of the automation but that's much faster than to create the backport itself and it will eventually go away.

- I agree... so this action is the only thing we can do with the resources we have
- I thought setting up a "Github App" reusing some apps that other teams might be using but I didn't have the chance to investigate this yet.

So what are your thoughts?
- Should we merge this PR and allow automatic PRs being created OR
- Should we expect every developer take care of the backports themselves?

carlosthe19916:
@gildub I am going to merge this PR and see how much helpful it is. There are a couple of PRs that need to be backported already

image

We can always revert this PRs if we find a better option.

PR #344, opened 2025-02-10
fix: Search Page does not search on the backend while typing

carlosthe19916:
Currently the Search Page does not search against the whole DB/Backend.

main branch:
-  the function `useAllEntities` fetches entities based on a `filterText` whose value is always empty `""`
- has a function `filterEntityListByValue` which is trimming the results of the backend.

After this PR:
- the function `useAllEntities` fetches continuously entities based on what the user types in the Input Text
- The function `filterEntityListByValue` has been removed and now we render "all" elements returned by the backend. The backend is better filtering data based on an input string so we do not need to cut/trim the original results.
- Although we do not filter the results of the backend, now with this PR we sort the entities list. Setting on top the entities that have a better match on the `title` and then `description` as a second alternative.
- Added a debounce to the Search Input text so we search in the backend using a delay of 500ms and avoid sending multiple requests.
- Enhanced the `ids` for entities to avoid collision (while running it I randomly saw that my browser told me multiple elements had the same id, not anymore with this change) 
```js
id: `advisory-${item.uuid}`,
```


client/src/app/pages/search/components/SearchMenu.tsx
```diff
@@ -133,24 +109,44 @@ function useAllEntities(filterText: string) {
...
+    id: `vulnerability-${item.identifier}`,
     title: item.identifier,
     description: item.description?.substring(0, 75),
     navLink: `/vulnerabilities/${item.identifier}`,
     type: "Vulnerability",
     typeColor: "orange",
   }));
 
-  tmpArray.push(
+  const filterTextLoweCase = filterText.toLowerCase();
```
kahboom, line 120: @carlosthe19916 Can you please change to `filterTextLowerCase`? Sorry but the typo will drive me crazy. ;D

Issue #149, opened 2024-09-04
Show the enabled importers at the top


jcrossley3:
I'm inclined to suggest we turn off pagination for the importers -- I can't imagine we'd ever have more than a dozen or so.

But regardless, can we show the enabled ones at the top, always? I often forget I have one enabled because I have to scroll to see it. And it's a PITA to scroll just to enable/disable one. 

I changed the backend to return the enabled ones first here: https://github.com/trustification/trustify/pull/759. But the UI overrides that and sorts by name, with no ability to sort by the "State" column.

jcrossley3:
@carlosthe19916 can you make this a priority for the next release, please? As someone working with importers, it's a major PITA 😄

carlosthe19916:
@jcrossley3 thanks for the reminder.

> I'm inclined to suggest we turn off pagination for the importers -- I can't imagine we'd ever have more than a dozen or so.

I think I can remove pagination and just show them all at one shoot. Or alternatively increase the default table size from 10 to 50.

> But regardless, can we show the enabled ones at the top, always? I often forget I have one enabled because I have to scroll to see it. And it's a PITA to scroll just to enable/disable one.

This one is very tricky even if it seems simple at first sight. Can you think of a similar app where you are able to run a task and then once running the element moves to the top?

Let me give you a real life scenario:

- There are 15 importers. All of them disabled.
- The user scrolls down the browser and goes to the last importer in the list/table. Eventually "clicks" on "run importer". ("enable importer" in our case). Then the importer start running.
  - If we follow your suggestion then the user is at the bottom of the page, click on "run" and then the row literally disappears from the vision of the user, because it was moved to the of the table/list.
  - So moving things actually creates a bigger problem.

The image below belongs to Jenkins and how you can click on the last right column of the list/table and run a task/job. When you run a task/job the elements do not move to the top just because they are running. Their position remains regardless of their status (running or not running). I am not saying Jenkins is the best UI but it behaves like that for the reasons I explained before.

![Image](https://github.com/user-attachments/assets/dc7cbc61-4723-4e6d-bafa-e8fd6a3f2d5e)

Don't get me wrong, I want to help you and reduce your PITA :) , but I would also like to avoid committing some UX mistakes.

WDYT?

- increasing the default size of the page to 50 would work?
- Removing the table and show all elements sounds better?
- OR could you share some UI from any mainstream app that manages "running tasks" and move things to the top so we can take it as an example?

jcrossley3:
[...]

PR #361, opened 2025-02-19
:sparkles:  add importer filters

carlosthe19916:
Related to #149

As expressed at https://github.com/trustification/trustify-ui/issues/149 , users find difficult to see the importers when they are too many.

This PR adds a filter to the importers table so users can define which status of importers they want to see.

https://github.com/user-attachments/assets/f5cbf1b4-d47f-41ab-bdec-4f693e9dd659

> further enhancements will be needed like the one proposed at https://github.com/trustification/trustify-ui/issues/149#issuecomment-2657181396 . But they can come in PR chunks and iterate over it.

### DIFF
--- a/client/src/app/pages/importer-list/importer-list.tsx
+++ b/client/src/app/pages/importer-list/importer-list.tsx
@@ -63,6 +63,21 @@ import { ANSICOLOR } from "@app/Constants";
 import { ImporterProgress } from "./components/importer-progress";
 import { ImporterStatusIcon } from "./components/importer-status-icon";
 
+type ImporterStatus = "enabled" | "disabled" | "scheduled" | "running";
+
+const getImporterStatus = (importer: Importer): ImporterStatus => {
+  const importerType = Object.keys(importer.configuration ?? {})[0];
+  const configValues = (importer.configuration as any)[
+    importerType
+  ] as SbomImporter;
+  const isImporterEnabled = configValues?.disabled === false;
+  if (!isImporterEnabled) {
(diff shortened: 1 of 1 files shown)


carlosthe19916:
@jcrossley3 What do you think about this?
I expect this not to fully solve your problem as it requires the user to set filters to visualize data.

I was planning to add the "notification" thing I mentioned at https://github.com/trustification/trustify-ui/issues/149#issuecomment-2657181396 to render running importers and their statuses. But I ran out of time and couldn't finish it so it will come in a separate PR.

Feel free to be brutally honest in case you don't like the filters :)

jcrossley3 (approved):
Sorry I missed this @carlosthe19916 !

You're correct: I don't think it goes far enough. It's better, though, so I'm happy to approve it, but I'd still like to see the running importers on that page at all times. 

So if you close #149 I may still open another issue in the future if I become annoyed enough. :stuck_out_tongue_winking_eye:

carlosthe19916:
@jcrossley3 thanks for the review. yeah, please feel free to open new issues. But give this one a try. 

I think in production, for v2.0, there are less than 5 default importers so I hope it won't be a big problem for them... but I still need to think about how to enhance the design of this page for better user experience.

jcrossley3:
> I think in production, for v2.0, there are less than 5 default importers so I hope it won't be a big problem for them...

They're lucky. Devs are forced to suffer: https://github.com/trustification/trustify/pull/1084

PR #365, opened 2025-02-22
:ghost: Rsbuild migration

carlosthe19916:
- This PR is using https://rsbuild.dev/guide/start/quick-start based on Rust
- rsbuild encourages to use [Biome](https://biomejs.dev/) (a Rust tool) to replace Eslint and Prettier. Biome just like Rsbuild is faster than Eslint and Prettier.
- Just like the other PR https://github.com/trustification/trustify-ui/pull/355 which explores Vite, this one explores Rsbuild as an alternative to replace our current webpack configuration.


carlosthe19916:
> @carlosthe19916, similarly as the Vite solution, it removes 21,000 lines of code and adds about 10,500 lines, even better than vite.
> 
> Did you use the migration option to convert from Webpack ?

No i didn't. I did all changes manually.
Important to notice that there are 2 tools:
- https://rsbuild.dev/
- https://rspack.dev/

Rsbuild contains a lot of "out of the box" configuration similar to Vite, and I did this POC with that.

carlosthe19916:
@gildub I made minor changes recently to this PR so it is now ready to be reviewed. Whenever you have time, feel free to look at it for the second time :)

client/src/app/components/AnalyticsProvider.tsx
```diff
@@ -25,6 +25,7 @@ export const AnalyticsProvider: React.FC<IAnalyticsProviderProps> = ({
 export const AnalyticsContextProvider: React.FC<IAnalyticsProviderProps> = ({
   children,
 }) => {
+  /* eslint-disable react-hooks/rules-of-hooks */
```
gildub, line 28: Biome doesn't use eslint stanza style.
Shouldn't that one be replace with `// biome-ignore lint/react-hooks-rules-of-hooks` ?

client/src/app/layout/header.tsx
```diff
@@ -41,6 +41,7 @@ export const HeaderApp: React.FC = () => {
     masthead: { leftBrand, leftTitle, rightBrand },
   } = useBranding();
 
+  /* eslint-disable react-hooks/rules-of-hooks */
```
gildub, line 44: Same here. Doesn't this to be replaced with Biome stanza ?

gildub (commented):
I wonder about the /* eslint-disable ... */ left in the code.
Unless I'm mistaken, don't we need to replace them with /* biome-disable ... */ equivalent ?

Besides above it looks really good and it's amazingly fast !!!

client/src/app/components/AnalyticsProvider.tsx
carlosthe19916: @gildub I removed the changes this PR was doing in regards of adding comments like `/* eslint-disable react-hooks/rules-of-hooks */` As you correctly mentioned it should be replaced by the Biome style.

We will need to do the migration in 2 steps:
- First step: this PR
- Second step: enable the biome lint and format rules in our CI and while doing it add the appropriate ``/* eslint-disable react-hooks/rules-of-hooks */`` or similar rules. At the moment we don't enable our Eslint rules in CI either.

gildub (approved):
LGTM

PR #368, opened 2025-02-23
:seedling:  remove mocks/browser from index.tsx

carlosthe19916:
It has been reported, internally, that some dependencies are being part of the final minimized bundles. E.g.

- If in the current main branch we execute `npm run build` and then `grep -r "myjino" client/dist/` we will see some results
- With the current PR we will avoid dependencies comming from `mocks/browser` being injected in the final bundles

We are not using the mocks/browser anyway. And it was originally intended to speed up the very first phase of development in the UI 1 year ago (when there was not backend therefore some mock api was needed to progress the UI).

### DIFF
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -4,7 +4,6 @@ import { createRoot } from "react-dom/client";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 
-import ENV from "@app/env";
 import App from "@app/App";
 import reportWebVitals from "@app/reportWebVitals";
 import "@app/dayjs";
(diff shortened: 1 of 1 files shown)


gildub:
@carlosthe19916 , that looks good meanwhile unless I'm wrong I can't start:dev as I get errors. Could you please double check ?

carlosthe19916:
> @carlosthe19916 , that looks good meanwhile unless I'm wrong I can't start:dev as I get errors. Could you please double check ?

What error are you seeing? I do see the error described at https://github.com/trustification/trustify-ui?tab=readme-ov-file#init-the-dev-server the very first time i start in a clean environment, but that is an error we had for a long time.

Besides that error I don't see anything. Could you share what error you see please?

https://github.com/user-attachments/assets/6d2d53b1-15d9-460e-a89d-eb5ebeed2f7e

carlosthe19916:
@gildub in case I didn't get a notification ↑↑... see my previous comment :)

carlosthe19916:
Just sharing here. I noticed that now that we don't push the hey-api (client) files to git, then if for any reason those files are not cleaned then the `npm run generate` seems not to overwrite them therefore generating an error related to the hey-api files. But that might be another bug not related to this PR.

gildub:
@carlosthe19916, thanks for pointing over the clean steps, I effectively had stalled files, likely the generated ones.

gildub (approved):

PR #374, opened 2025-02-25
:sparkles: Stop fetching data every X seconds and do it only when requested 

carlosthe19916:
Fixes: https://github.com/trustification/trustify-ui/issues/325

Removes all the re-fetch logic with the exception of the Importers as we still need to follow the progress of an importer

- Moves the fetch interval value to a single place (Contants)
- This way we can change the intervals at one shot in a single place

### DIFF
--- a/client/src/app/Constants.ts
+++ b/client/src/app/Constants.ts
@@ -6,6 +6,8 @@ export const RENDER_DATE_FORMAT = "MMM DD, YYYY";
 export const RENDER_DATETIME_FORMAT = "MMM DD, YYYY | HH:mm:ss";
 export const FILTER_DATE_FORMAT = "YYYY-MM-DD";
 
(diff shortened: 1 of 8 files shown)


gildub:
@carlosthe19916, I started the review but I'd like to do more tests if you don't mind.

gildub:
@carlosthe19916, The title is misleading, this PR changes the refetch interval to the default which is zero.
I'm saying because the refetch "logic" is still accessible by calling refetch.
Actually I wonder why the refetch is not used anywhere withing the application.
Maybe you used a different trick ?

gildub:
@carlosthe19916, thanks for changing the title, that's a detail, the main thing being for me to understand the evolution of the project.

gildub (approved):

PR #379, opened 2025-02-26
:sparkles: sbom details page - filter out duplicate packages

carlosthe19916:
Fixes https://github.com/trustification/trustify-ui/issues/369

The SBOM Details page / Vulnerability Tab contains duplicate packages. This PR just filters them out.

To compare the main Results vs this PR results we can import the 2 files attached to the issue https://github.com/trustification/trustify-ui/issues/369 and then go to the "The SBOM Details page / Vulnerability Tab"

This is what I see with this PR 
image

Note that there are 4 libnghttp2 but all of them belong to a different purl (you can see they have different architecture)


dejanb (approved):
Yeah, the 4 packages with different qualifiers is the correct result

PR #383, opened 2025-02-27
:sparkles: Add 'Unknown' Severity when severity=null

carlosthe19916:
Workaround for https://github.com/trustification/trustify/issues/1362

Fixes: https://github.com/trustification/trustify-ui/issues/359 , https://github.com/trustification/trustify-ui/issues/358 and https://github.com/trustification/trustify-ui/issues/357

For Typescript and the current REST Endpoint. Severities can be of type `Severity | null`. So far we were just ignoring Severities whose values were `null`, not anymore with this PR as `null` values will become `Unknown`.

So we have 6 types of Severity now:
- Unknown => fake one as it is dynamically handled by the UI
- None
- Low
- Medium
- High
- Critical

All types of severities were added to Barcharts, Donut Charts, Filters, Sort, etc.

- Also taking the opportunity to delete 2 files that are not used anywhere.

image

Screenshot From 2025-02-27 14-55-10

Screenshot From 2025-02-27 14-54-58

Screenshot From 2025-02-27 14-54-46


gildub (approved):
LGTM

The ExtendedSevery type is perfect to handle this !

PR #385, opened 2025-03-03
:sparkles: Set default sorting to main tables

carlosthe19916:
Fixes https://github.com/trustification/trustify-ui/issues/116

Set default sorting to always see the latest entities in the table.

- Advisory table sorted by "modified" date by default
- Sbom table sorted by "published" date by default
- Vulnerability sorted by "published" date by default


gildub (approved):
Technically looks good meanwhile it's weird to have objects sorted in way the user can't see it as there is no published date (for instance for Sboms) displayed.

carlosthe19916:
> Technically looks good meanwhile it's weird to have objects sorted in way the user can't see it as there is no published date (for instance for Sboms) displayed.

Thanks for the approval.

The `published` is the "Created on" column in rendered in the page

Screenshot From 2025-03-06 10-33-08

Screenshot From 2025-03-06 10-35-05

PR #408, opened 2025-03-21
:ghost:  Add Biome Format Rules

carlosthe19916:
After the recent migration to RsBuild we need to make iterations to start applying Biome best practices

This PR Enables the format check in CI and applies all the formats required by Biome.

It is the result of executing `npm run format` locally


carlosthe19916:
Merging this PR as it contains only format changes and it does not touch in any way the logic or any component. Also all the tests passed successfully.

PR #447, opened 2025-04-03
:bug: [TC-2397] Run should only appear if importer is scheduled

carlosthe19916:
https://issues.redhat.com/browse/TC-2397

An Importer can have only 3 statuses:

- "disabled"
- "scheduled" -> means it is enabled but waiting to be executed
- "running"

This will allow the RUN action to appear only when the Importer is scheduled but disappear when the Importer is either running or disabled

### DIFF
--- a/client/src/app/pages/importer-list/importer-list.tsx
+++ b/client/src/app/pages/importer-list/importer-list.tsx
@@ -104,7 +104,7 @@ export const ImporterList: React.FC = () => {
   };
 
   const { mutate: updateImporter } = useUpdateImporterMutation(
-    () => { },
+    () => {},
     onEnableDisableError,
   );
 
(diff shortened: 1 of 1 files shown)


carlosthe19916:
That is a fair point. We currently have 3 action buttons: "Enable" | "Disable" | "Run". 
- If the Importer is disabled then only the button "Enable" appears
- If the Importer is enabled but waiting, only the "Run" and "Disable" button appear
- If the Importer is running the only the "Disable" button appear

So at the moment we rely on "Hiding" and "Showing".

If we go to the approach of "disabling" a button I wonder if that might bring inconsistencies.
- We always rely on "enabling|disabling" actions OR
- we always "hide|render" actions

WDYT? I don't know what would be the most practical decision

carlosthe19916:
@gildub a follow up. I've just looked at the Patternfly library and it is not possible to "enable/disable" action buttons so the current approach of hiding/showing action buttons is the only way.

What is possible is to disable the kebab dropdown of the actions but that will disable all actions and not just one in particular.

Do you know how to disable single actions in a table using the Patternfly library? Perhaps that might help

gildub:
@carlosthe19916 , 

From experience, the menu options are not dynamic, by that I mean they do not change depending on the context.
Options don't disappear but disabled, there is no escape from that.

Regarding PF, I remember that it's tricky to disable en entry directly meanwhile there are ways to work around that and implement.

Let me know if you would like me to take care of it.

carlosthe19916:
@gildub you were right. My apologies for missing some information before. Now the "Run" action is enabled/disabled

gildub (approved):

PR #454, opened 2025-04-08
Runtine : Use Bun

gildub:
This is  a POC

The application works just fine but ci doesn't because still using node.js


client/package.json
```diff
@@ -5,20 +5,21 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "clean": "rimraf ./dist",
-    "clean:all": "rimraf ./dist ./node_modules",
-    "build": "npm run generate && rsbuild build",
+    "clean": "bunx rimraf ./dist",
+    "clean:all": "bunx rimraf ./dist ./node_modules",
```
RiskyMH, line 9: btw you can just use `rm -rf` with bun as on windows it emulates some posix commands

gildub: @RiskyMH, thanks for the suggestion but could you please clarify ? (Sorry for delayed response)

RiskyMH: On windows, bun uses its own shell to run the scripts. This means the items in https://bun.sh/docs/runtime/shell#builtin-commands work in bun. So instead of `bunx rimraf ...`, you can use the standard and faster `rm -rf ..`.

The reason I mention windows is that the other bun targets are posix which mean they implement the rm commands and co.

PR #482, opened 2025-05-05
:sparkling: WIP SBOM|Advisory labels

carlosthe19916:
Add Advisory and SBOM Labels

### DIFF
--- a/client/src/app/components/EditLabelsForm.tsx
+++ b/client/src/app/components/EditLabelsForm.tsx
@@ -1,0 +1,163 @@
+import React from "react";
+
+import {
+  ActionGroup,
+  Button,
+  ButtonVariant,
+  Form,
+  FormGroup,
+  Grid,
+  GridItem,
+  Label,
+  LabelGroup,
+  Stack,
+  StackItem,
+  TextInput,
+} from "@patternfly/react-core";
+
+interface EditLabelsFormProps {
+  title: string;
+  value: { [key: string]: string };
+  isDisabled: boolean;
+  onSave: (value: { [key: string]: string }) => void;
+  onClose: () => void;
+}
+
+export const EditLabelsForm: React.FC<EditLabelsFormProps> = ({
+  title,
+  value,
+  isDisabled,
+  onSave,
+  onClose,
+}) => {
+  const [inputKey, setInputKey] = React.useState("");
+  const [inputValue, setInputValue] = React.useState("");
+
+  const [labels, setLabels] = React.useState(
+    Object.entries(value).sort(([keyA], [keyB]) => keyB.localeCompare(keyA)),
+  );
+
+  const addLabel = () => {
+    if (!inputKey) {
+      return;
+    }
+
+    const newLabels = labels.filter(([key]) => key !== inputKey);
+    setLabels([...newLabels, [inputKey, inputValue]]);
+
+    setInputKey("");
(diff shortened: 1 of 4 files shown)


gildub:
@carlosthe19916,

This has nothing to do with with patch but be aware that Codecov makes it impossible to review the code on Github.

Regarding the Labels, 

My understanding is that the user shouldn't be able to edit/delete internal labels like `type=SPDX`.

The Edit Labels window has text flickering (it's very fast and barely visible) above the `save` button.

Adding Labels is not possible.

carlosthe19916:
> @carlosthe19916,
> 
> This has nothing to do with with patch but be aware that Codecov makes it impossible to review the code on Github.

I agree. I think we could move the codevov thing only to commits in branches and not in PRs.
We could always have tests here but I think our test efforts will be mainly in the trustify-tests repository.

I'll prepare a new PR to do that.

> Regarding the Labels,
> 
> My understanding is that the user shouldn't be able to edit/delete internal labels like `type=SPDX`.

Currently there is nothing that prevents any user to edit/delete/create labels as the REST API allows those operations. It has been agreed that it can be an improvement for future release/iterations. The lock for preventing such operations needs to be done first in the backend and then in the UI.

> The Edit Labels window has text flickering (it's very fast and barely visible) above the `save` button.
> 

As per design, the list of labels are fetched from the backend according to what the user types in the InputText

So the flickering would be expected as new data is coming from the backend and replace the Dropdown elements.

> Adding Labels is not possible.
Uhm... I would need to verify it but I am sure I tested it and it worked. 
- A label can be added pressing "Enter" or selecting an element of the dropdown

carlosthe19916:
> @carlosthe19916,
> 
> If the UI cannot prevent user to shoot him/her self in the foot then I believe we need to wait for the backend to be ready so the UI can support the feature properly.
> 

This is out of our hands, it has been decided that we need Labels for the release and for the time being users will be able to apply CRUD operations to all labels, including the `type` labels.

> Regarding the flickering effect I'm not sure we're talking about the same thing, let me capture that so you can see it. The text "Loading" appears briefly then disapears:
> 
>  Screencast.From.2025-06-10.14-26-49.mp4 

Thanks for pointing that out, I added a commit that should fix the issue.

> Regarding addition of new labels : Still not working or I totally missed how it should work ;)

You could add labels pressing Enter or selecting any of the elements of the dropdown. I added a new element in the dropdown "Create a new option" to make it obvious that you could create new labels even if the are not part of the original dropdown. That is in sync with the patternfly demo pages like https://www.patternfly.org/components/menus/select#typeahead-with-create-option


https://github.com/user-attachments/assets/98396a89-f4bc-472f-8d56-f5fa5c47d681

gildub (commented):
Everything works now as expected.
The only concern I have is about naming interface `SingleLabel`.
This looks like a wrong name where `Label` would suffice.
I understand a conflict might exist `Labels` type provided by OpenAPI but that should be resolved with an alias.

carlosthe19916:
> Everything works now as expected. The only concern I have is about naming interface `SingleLabel`. This looks like a wrong name where `Label` would suffice. I understand a conflict might exist `Labels` type provided by OpenAPI but that should be resolved with an alias.

As agreed internally I renamed SingleLabel to Label

gildub (approved):

PR #592, opened 2025-07-09
Labels : Fix create new label

gildub:
Make sure the selection is disabled after clicking a select option.

Fix https://issues.redhat.com/browse/TC-2661


crate/build.rs
```diff
@@ -17,7 +17,7 @@ static NPM_CMD: &str = "npm";
 fn main() {
     println!("Build Trustify - build.rs!");
 
-    println!("cargo:rerun-if-changed={}", UI_DIR_SRC);
+    println!("cargo:rerun-if-changed={UI_DIR_SRC}");
```
carlosthe19916, line 20: This change is not needed

gildub: @carlosthe19916, if you want I could push that one in a separate PR as it's not related to this issue.
What do you think ?

carlosthe19916 (approved):

PR #632, opened 2025-08-04
Bump React to version 19

gildub:
Resolves https://github.com/trustification/trustify-ui/issues/329

## Summary by Sourcery

Upgrade core frontend dependencies to support React 19

Enhancements:
- Bump React and React DOM to version 19.1.1
- Upgrade React Router DOM to version 7.7.1
- Update PatternFly packages to their 6.3.x and 8.3.x releases


carlosthe19916:
On a side note. If by any change we discover that the root of the issue is the upgrade of `react-router-som` then we could evaluate if we actually need to upgrade it.
- If this package does not need to be upgraded we can keep the same version and address that upgrade in a separate PR
- If this package is needed then we need to find out what to do to avoid slow navigation

carlosthe19916:
> @carlosthe19916, to be on the safe side, let's double check we aren't experiencing such navigation slow down behavior.

I agree, give me some time to test it again... I think the problem was that for some reason when you initially upgraded react-router-dom then the Spinner, that is usually rendered between switching pages while the browser if fetching static files from the server, was not rendered. That was one concrete thing I noticed. I guess that package deserve its own PR and analysis.

carlosthe19916 (approved):
LGTM! Thanks @gildub . I confirm, the slow navigation while switching pages is fixed.

- Please add the label to backport it to the release/0.3.z branch and merge it when you feel it is the right time

PR #655, opened 2025-08-14
:seedling: add date filter and replace dateRange

carlosthe19916:
Fixes: https://github.com/trustification/trustify-ui/issues/393, https://github.com/trustification/trustify-ui/issues/394, https://issues.redhat.com/browse/TC-2393, and https://issues.redhat.com/browse/TC-2402

- This PR creates a new FilterType `date`.
- All current filters using "dateRange" are being replaced by "date". This way we avoid all issues described in the JIRAs and Github issues linked above.

### DIFF
--- a/client/src/app/components/FilterPanel/DateFilter.tsx
+++ b/client/src/app/components/FilterPanel/DateFilter.tsx
@@ -1,0 +1,61 @@
+import React, { type FormEvent, useState } from "react";
+
+import { DatePicker, Form, FormGroup } from "@patternfly/react-core";
+
+import {
+  americanDateFormat,
+  isValidAmericanShortDate,
+  parseAmericanDate,
+} from "../FilterToolbar/dateUtils";
+import type { IFilterControlProps } from "./FilterControl";
+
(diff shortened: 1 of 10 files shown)


gildub (commented):
Not sure if that was covered in the UXD but what about the date format depending on the locale ?

gildub:
@carlosthe19916,

Yes, I think we need to circle that back with UXD.

From a functional viewpoint, we have :

- Use case 1 
A date filter is needed to filter items by creation date, the user can choose a date, and decide to filter items before or after that date.

- Use case 2
A date filter is needed to select items by creation date using a range of dates.

Use case 1 could be implemented in different ways and that's where all the troubles came from.
Using a couple of date filters displaying before/after (to replace the confusing from/to) is easily confusing because it looks like a range (and cannot behave as a range logically).
That could be mitigated by making it explicit to the user it's not a range.
The third solution I mentioned is to have (if possible with the table hook filter, I haven't verified) a `smart` customized date filter in one single field where the user can choose before/after (or the carets > or <) without ambiguity.

The requirements must decide when we need either use cases.

PR #680, opened 2025-08-26
ci: biome should fail on warnings

carlosthe19916:


carlosthe19916:
@mrrajan This PR should avoid having warnings on the Biome/Typescript check process

PR #697, opened 2025-09-01
tests: Add Advisory UI tests

vobratil:
A set of tests that test basic Advisory Explorer UI functionality, migrated from the trustify-tests repository.

### DIFF
--- a/e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
+++ b/e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
@@ -2,6 +2,62 @@ Feature: Advisory Explorer
     Background: Authentication
         Given User is authenticated
 
+# Search for advisories
+    Scenario: Search for an advisory using the general search bar
+        
+        When User searches for an advisory named "<advisoryID>" in the general search bar
+        
+        Then The advisory "<advisoryID>" shows in the results
+
+        Examples:
+            | advisoryID      |
+            | CVE-2024-26308  |
+    
+    Scenario: Search for an advisory using the dedicated search bar
+
+        When User searches for "<advisoryID>" in the dedicated search bar
+        
+        Then The advisory "<advisoryID>" shows in the results
+
+        Examples:
+            | advisoryID      |
+            | CVE-2024-26308  |
+
+    # Advisory Explorer
+    Scenario: Display an overview of an advisory
+
+        When User visits Advisory details Page of "<advisoryID>"
+
+        Then The page title is "<advisoryID>"
+        Then The "Download" button is visible
+
+        Examples:
+            | advisoryID      |
+            | CVE-2024-26308  |
+
+    Scenario: Download an advisory
+
+        When User visits Advisory details Page of "<advisoryID>"
+        And User clicks the "Download" button
+
+        Then File with the name "<fileName>" is downloaded
+
+        Examples:
+            | advisoryID      | fileName            |
+            | CVE-2024-26308  | CVE-2024-26308.json |
+
+    Scenario: Display the Info tab
+
+        When User visits Advisory details Page of "<advisoryID>"
+
+        Then The "Overview" panel is visible
+        Then The "Publisher" panel is visible
+        Then The "Tracking" panel is visible
+
+        Examples:
+            | advisoryID      |
+            | CVE-2024-26308  |
+
 # Advisory Vulnerabilities
 Scenario: Display vulnerabilities tied to a single advisory
     Given User visits Advisory details Page of "<advisoryName>" with type "<advisoryType>"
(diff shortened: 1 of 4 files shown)


e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
```diff
@@ -2,6 +2,62 @@ Feature: Advisory Explorer
...
+            | advisoryID      |
+            | CVE-2024-26308  |
+
+    # Advisory Explorer
+    Scenario: Display an overview of an advisory
+
+        When User visits Advisory details Page of "<advisoryID>"
+
+        Then The page title is "<advisoryID>"
+        Then The "Download" action is available
```
queria, line 32:
```suggestion
        And The "Download" action is available
```

e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
```diff
@@ -2,6 +2,62 @@ Feature: Advisory Explorer
...
+            | advisoryID      | fileName            |
+            | CVE-2024-26308  | CVE-2024-26308.json |
+
+    Scenario: Display the Info tab
+
+        When User visits Advisory details Page of "<advisoryID>"
+
+        Then The "Overview" panel is visible
+        Then The "Publisher" panel is visible
+        Then The "Tracking" panel is visible
```
queria, line 55:
```suggestion
        And The "Tracking" panel is visible
```

e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
queria, line 32: Maybe the 'action is available' is just redundant?
Given there is just below test which does need this action to be available since it is verifying its effect?

It could be usefull only if we have actions which we do not execute - when their functionality is not covered with explicit test (yet or for whatever other reason).

queria (dismissed):

queria (commented):
Actually looking at existing code there is mixture of Then-Then or Then-And usage,
and so currently it does not matter which style is used in this PR.

e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
```diff
@@ -2,6 +2,62 @@ Feature: Advisory Explorer
...
+    Scenario: Download an advisory
+
+        When User visits Advisory details Page of "<advisoryID>"
+        And User selects the "Download Advisory" action
+
+        Then File with the name "<fileName>" is downloaded
+
+        Examples:
+            | advisoryID      | fileName            |
+            | CVE-2024-26308  | CVE-2024-26308.json |
```
carlosthe19916, line 47: @vobratil this Scenario will fail in the majority of executions. I would like us to rewrite the implementation of this whole scenario please.

### Why it fails?
- `And User selects the "Download Advisory" action` clicks on the Download Button
- `File with the name "<fileName>" is downloaded` actives a "watcher" to wait for download actions in the browser. E.g. `const downloadPromise = page.waitForEvent("download");`
  - Here is the problem, because by the time the "watcher" is enabled, the Download Button has been clicked already and the event triggered and not catched

### How it should be?
- Before clicking in the Download Button, we need to enable the "watcher"
- Click on the Button
- Catch the Download event

Long story story short:
```
// Create watcher to wait for events
const downloadPromise = page.waitForEvent("download");

// Write code for clicking in the Download Button
const detailsPage = new DetailsPage(page);
await detailsPage.clickOnPageAction("Download Advisory");

// Catch Download event and do the assertions we might need
const download = await downloadPromise;
const actualFilename = download.suggestedFilename();
expect(actualFilename).toEqual("CVE-2024-26308.json");
```

carlosthe19916 (changes requested):
@vobratil Thanks for the contribution. And sorry for my late review.
There is only one major thing I would like this PR to change, see my comment about refactoring the `Scenario: Download an advisory` scenario. It is failing, and even the Code Coverage test is alerting us about it.

e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
vobratil: @carlosthe19916 Thanks! I've changed it to use a method that somebody else implemented for the SBOM Explorer, so it should be okay now. It's kind of bending the way those feature files are supposed to work, but I guess that there is no way around it in this case.

e2e/tests/ui/features/@advisory-explorer/advisory-explorer.feature
vobratil: @queria I agree that it's kind of useless here now, but I'd leave it in case we have more actions in the future.

vobratil:
@queria I've changed the download test according to your suggestion and all the tests are passing now. If you think the PR is ready now, I'd be grateful for an approval :).

carlosthe19916 (approved):
@vobratil LGTM! Thanks, merge it whenever you are ready.
Please don't forget to backport this PR to the branch 0.3.z. Just add the backport label before merging the PR

PR #698, opened 2025-09-01
ci: reference trustify-release-tools

carlosthe19916:
This PR moves all old references from `trustification/release-tools` to `guacsec/trustify-release-tools`


carlosthe19916:
The backport bot is not working yet, so adding the "needs-backport" for when we are able to backport things

PR #701, opened 2025-09-03
feat: Scan Report + Static Report

carlosthe19916:
This PR contains:

- `client/src/static-report` directory, which contains a standalone set of files for generating the SCAN SBOM report
  - Relies on its own `client/rsbuild.static-report.config.ts` configuration file, because it can be built and generate its own JS and CSS files.
  - running `npm run build` should generate `client/dist` + `client/dist/static-report`, both contain the minified JS and CSS static files
- For being able to use `useBlocker` from `react-router-dom` it was necessary to migrate from `BrowserRouter` to `RouterProvider`

This PR depends on https://github.com/guacsec/trustify/pull/1952 and we need both PRs for them to work.

Also here is a demo of how it will look like:

https://github.com/user-attachments/assets/e53f2b25-e6d8-4b31-afe0-05c97020610d


carlosthe19916:
replaced by https://github.com/guacsec/trustify-ui/pull/725

PR #704, opened 2025-09-04
fix: hide upload btn on SearchPage/SbomTab

carlosthe19916:
The Button for uploading SBOM should not be visible in the SEARCH PAGE but only on the SBOM List page.

The current main branch looks like the image below:

image (3)

This PR avoids rendering that button on the SEARCH PAGE

### DIFF
--- a/client/src/app/pages/sbom-list/sbom-toolbar.tsx
+++ b/client/src/app/pages/sbom-list/sbom-toolbar.tsx
@@ -35,12 +35,19 @@ export const SbomToolbar: React.FC<SbomToolbarProps> = ({ showFilters }) => {
   return (
     <Toolbar {...toolbarProps} aria-label="sbom-toolbar">
       <ToolbarContent>
-        {showFilters && <FilterToolbar {...filterToolbarProps} />}
-        <ToolbarItem>
-          <Button variant="primary" onClick={() => navigate(Paths.sbomUpload)}>
-            Upload SBOM
-          </Button>
-        </ToolbarItem>
+        {showFilters && (
+          <>
+            <FilterToolbar {...filterToolbarProps} />
+            <ToolbarItem>
+              <Button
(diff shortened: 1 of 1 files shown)


gildub (commented):
@carlosthe19916, I might be missing something but what does actually make the Upload button disappear ?

carlosthe19916:
> @carlosthe19916, I might be missing something but what does actually make the Upload button disappear ?

@gildub the button, in this PR, is part of the conditional rendering defined by `showFilters`.

client/src/app/pages/sbom-list/sbom-toolbar.tsx
```diff
@@ -35,12 +35,19 @@ export const SbomToolbar: React.FC<SbomToolbarProps> = ({ showFilters }) => {
   return (
     <Toolbar {...toolbarProps} aria-label="sbom-toolbar">
       <ToolbarContent>
-        {showFilters && <FilterToolbar {...filterToolbarProps} />}
-        <ToolbarItem>
-          <Button variant="primary" onClick={() => navigate(Paths.sbomUpload)}>
-            Upload SBOM
-          </Button>
-        </ToolbarItem>
+        {showFilters && (
```
gildub, line 38: It would make sense to have a showUploadButton or equivalent in order to follow the same logic as showFilters. Alternatively the upload button could be separate from the SbomToolbar.

gildub: @carlosthe19916, 

If the toolbar contains features such as filter, action buttons and pagination then the optional features need to have a conditional to enable/disable them. 
It would therefore make sense to have a component property to control each optional feature.
I believe we have to keep in mind that we might want to refactor that one day and make the toolbar even more generic, to handle sbom, advisories, etc.

gildub (approved):

PR #714, opened 2025-09-09
SBOM Deletion Performance Test

vobratil:
A simple test that measures the response times on deletion of SBOMs. I also added two request and response interceptors to the fixtures to make the response time measuring possible.

### DIFF
--- a/e2e/tests/api/features/performance.ts
+++ b/e2e/tests/api/features/performance.ts
@@ -1,0 +1,8 @@
+import { test } from "../fixtures";
+
+// test.beforeEach()
+
+test("@performance Delete / Large SBOMs", async ({ axios }) => {
+  
+    
+});
--- a/e2e/tests/api/fixtures.ts
+++ b/e2e/tests/api/fixtures.ts
@@ -95,6 +95,7 @@ const initAxiosInstance = async (
   access_token = tokenResponse.access_token;
 
   // Intercept Requests
+  // Add access token
   axiosInstance.interceptors.request.use(
     (config) => {
       config.headers.Authorization = `Bearer ${access_token}`;
@@ -106,7 +107,16 @@ const initAxiosInstance = async (
     },
   );
 
+  // Measure request start time
+  axiosInstance.interceptors.request.use(
+    (config) => {
+      config.headers['request-startTime'] = new Date().getTime();
+      return config
+    }
+  );
+
   // Intercept Responses
+  // Retry
   axiosInstance.interceptors.response.use(
     (response) => {
       return response;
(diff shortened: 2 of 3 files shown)


e2e/tests/api/fixtures.ts
```diff
@@ -106,7 +107,16 @@ const initAxiosInstance = async (
     },
   );
 
+  // Measure request start time
+  axiosInstance.interceptors.request.use(
+    (config) => {
+      config.headers['request-startTime'] = new Date().getTime();
```
queria, line 113: I would prefer to not use request headers (which are sent to server/backend api) for storing timing information.

But seems on first look there is not an easy alternative available.

While Axios does not restrict/complain about extra properties on config object (and seems its meant to be enriched with extra fields by interceptors),
in typescript it is not possible to inject custom properties into objects.

So the proper clean way seems would be to augment the module object types definitions, e.g.:
```typescript
import 'axios';

declare module 'axios' {
  export interface AxiosRequestConfig {
    sentAt?: number;
  }
  export interface AxiosResponse {
    duration?: number;
  }
}
```
this would enable here to store it as
```typescript
config.sentAt = new Date.now();
```
and then in response interception
```typescript
response.duration = new Date.now() - response.config.sentAt;
```

Although, it may be ok for now to keep storing it in headers too, it is just not ideal as it unnecessarily gets sent to the server, and in rare cases could possibly lead to issues (if server would care about such/extra headers).

e2e/tests/api/features/performance.ts
```diff
@@ -0,0 +1,114 @@
...
+      .delete(`/api/v2/sbom/${sbomId}`)
+      .then((response) =>
+        writeRequestDurationToFile(
+          reportFile,
+          "n/a",
+          response.data.id,
+          String(response.duration),
+        ),
+      )
+      .catch((error) => {
```
carlosthe19916, line 89: We could also record the failure as part of the report and not just log

e2e/tests/api/helpers/delete.ts
```diff
@@ -0,0 +1,30 @@
...
+  var existingSbomIds = [];
+
+  for (const sbomId of sbomIds) {
+    try {
+      await axios.get(`/api/v2/sbom/${sbomId}`);
+      existingSbomIds.push(sbomId);
+    } catch (_error) {
+      logger.info(`SBOM with ID ${sbomId} does not exist anymore. Skipping.`);
+    }
+  }
```
carlosthe19916, line 15: I think this is not necessary... since it is only cleaning we can call directly `axios.delete`, if the sbom does not exist, it will fail and nothing happens

e2e/tests/api/helpers/delete.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const deletionPromises = existingSbomIds.map((sbomId) =>
+    axios.delete(`/api/v2/sbom/${sbomId}`).catch((error) => {
+      logger.error(`Failed to delete SBOM with ID: ${sbomId}`, error);
+    }),
```
carlosthe19916, line 20: As this is a helper function, helpers should not handle errors but let the caller of the function to deal with the exceptions. Handling the catch here means that if it fails we will only see a log but the function will succeed giving the false impression that the function was executed correctly.

carlosthe19916 (changes requested):
@vobratil sorry for my late review. It took me some time to do it. Please read my comments below.

It is a nice addition of tests, well done. Just some minor code enhancements I am suggesting.

Let me know if you need me to expand on any of the points mentioned in my review

PR #715, opened 2025-09-09
Add automation to importers UI

ikanias:
This PR contains the .feature file and the .ts file of importer-explorer for the importers UI automation in TPA.

## Summary by Sourcery

Add end-to-end UI automation for the Importer Explorer in TPA, including Playwright BDD step definitions and feature scenarios covering navigation, filtering, search, row expansion, pagination, and importer action workflows.

Tests:
- Add Playwright BDD step definitions for the Importer Explorer UI
- Add importer-explorer.feature covering navigation, filtering, search, row expansion, pagination, and action menu scenarios
- Add tests to verify table column headers, filter icon, and search results
- Add tests to expand importer rows and validate sub-tables, pagination states, and importer statuses
- Add tests to assert pagination controls disablement and non-editability of page inputs
- Add tests to verify Call to Action menu options for running, enabling, and disabling importers with confirmation dialogs and status updates

### DIFF
--- a/e2e/tests/ui/features/@importer-explorer/importer-explorer.feature
+++ b/e2e/tests/ui/features/@importer-explorer/importer-explorer.feature
@@ -1,0 +1,78 @@
+Feature: Importer Explorer - View importers details
+    Background: Authentication
+        Given User is authenticated
+
+    Scenario: The user navigates to Importers Explorer page
+        Given The user navigates to TPA URL page successfully
+        When The user selects the importers menu option
+        Then Application navigates to Importers page
+        Then Call to Action button for each importer should be visible
+
+    Scenario: The user can see that the filter icon appears in importers page
+        Given The user navigates to TPA URL page 
+        When The user selects the importers overview page
+        Then The column headers should be visible
+        Then There should be a filter icon displayed
+
+    Scenario: The user can search a text in search text box with pressing 'Enter' in the keyboard 
+        Given The user navigates to TPA URL landing page
+        When The user selects the importers option
+        Then A search text box appears
+        Then The user types "<importerName>" in the search text box and presses Enter key
+        Then Importer total results is 1
+    
+        Examples:
+            | importerName |
+            | cve |
+    
+    Scenario: The user can expand all importers to see their logs
+        Given The user navigates to TPA URL dashboard page
+        When The importers option is selected on the left menu
+        Then The user expands all the importers rows
+        Then All the importers rows should be expanded
+        
+    Scenario: Pagination controls on importer explorer page
+        Given The user navigates to TPA URL page home page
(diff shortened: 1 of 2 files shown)


e2e/tests/ui/features/@importer-explorer/importer-explorer.feature
```diff
@@ -0,0 +1,78 @@
+Feature: Importer Explorer - View importers details
+    Background: Authentication
+        Given User is authenticated
+
+    Scenario: The user navigates to Importers Explorer page
+        Given The user navigates to TPA URL page successfully
```
queria, line 6: I think it is better to avoid these extra 'navigate to homepage' generic steps. And especially repeating them in every scenario too.

- You already have `Background` - `Given User is authenticated` (similar to existing advisory/sbom/vulnerability-explorer tests)
-- That can be expected to result in You being already on TPA website (so have left menu available)

Also if similar step would be needed, then:
- If all `Scenarios` under the `Feature` need such step it would be better to add it to `Background` instead
- The TPA URL page - You try to navigate to Dashboard, we should in current situation avoid navigation to Dashboard unless it is relevant for the test (e.g. of dashboard itself), as currently it is very expensive page (compared to most), so if then it is better for these universal-not-related-to-test cases better to navigate elsewhere like to `/upload` or `/importers` which are way faster and do not slow test execution that much, as also do not cause that big load on the backend

e2e/tests/ui/features/@importer-explorer/importer-explorer.step.ts
```diff
@@ -0,0 +1,449 @@
+import { createBdd } from "playwright-bdd";
+import { ElementHandle, expect } from "playwright/test";
+import { DetailsPage } from "../../helpers/DetailsPage";
+import { ToolbarTable } from "../../helpers/ToolbarTable";
+import { SearchPage } from "../../helpers/SearchPage";
+import { test } from "@playwright/test";
```
queria, line 6: We have tweaked/extended 'test' object, so please replace with:
```suggestion
import { test } from "../../fixtures";
```

carlosthe19916:
@ikanias @queria any progress on this PR? If you agree, let's switch it to "Draft" status until all github ci workflows pass successfully.

vobratil (changes requested):
@ikanias Unfortunately this PR has quite a lot of problems with things like defining redundant test steps, not taking advantage of already defined methods, relying heavily on very specific xpaths, unnecessary imports and others. I can suggest taking a look at how other user stories and their steps are defined, using the [Playwright Code Gen tool](https://playwright.dev/docs/codegen) to help create maintainable code and perhaps even take advantage of the Trustify MCP server to generate good test steps from the BDD file.

PR #722, opened 2025-09-11
task: enhance devcontainer cucumber settings

carlosthe19916:
This change will allow to jump from the .feature file straight to the Typescrip definition of the Step. Only valid for VSCode users


vobratil (approved):
LGTM!

PR #729, opened 2025-09-14
tests (fix): enhance table wait assertions to avoid flaky tests

carlosthe19916:
Code enhancements to wait helpers.
- Introduction of "poll" to recurrently verify data until the expected assertion passes.
- Enhance `waitUntilDataIsLoaded` => adding a reasonable time to wait for the table start transitioning from "loading" to "loaded" before the assertions are done.

### DIFF
--- a/e2e/tests/ui/pages/Table.ts
+++ b/e2e/tests/ui/pages/Table.ts
@@ -23,14 +23,19 @@ export class Table {
     return result;
   }
 
+  /**
+   * Waits until the table is not in "Loading" state.
+   */
   async waitUntilDataIsLoaded() {
+    // If table already has data, we need to wait to be sure we are not asserting the previous page
+    await this._page.waitForTimeout(500);
+
     const rows = this._table.locator(
       'xpath=//tbody[not(@aria-label="Table loading")]',
     );
     await expect(rows.first()).toBeVisible();
 
-    const rowsCount = await rows.count();
-    expect(rowsCount).toBeGreaterThanOrEqual(1);
+    await expect.poll(() => rows.count()).toBeGreaterThanOrEqual(1);
   }
 
   async clickSortBy(columnName: string) {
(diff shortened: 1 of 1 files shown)


e2e/tests/ui/pages/Table.ts
```diff
@@ -23,14 +23,18 @@ export class Table {
     return result;
   }
 
-  async waitUntilDataIsLoaded() {
+  /**
+   * @param waitMs - Optional. Milliseconds to wait before checking table data.
+   */
+  async waitUntilDataIsLoaded(waitMs: number = 500) {
```
vobratil, line 29: Could we please move things like this to some kind of config file? I'm not sure if the Playwright config file is suitable, but this value may need to be changed on different environments or in specific conditions and it would be great to have them all in one place, easily changeable, instead of having to dive into the code and changing a hardcoded value.

vobratil (changes requested):
I would like to ask for the waiting period to be moved to a separate file where we can configure it along with any other such values. It would be great if we could avoid values hardcoded in the tests and helper classes.

e2e/tests/ui/pages/Table.ts
carlosthe19916: @vobratil the function `waitUntilDataIsLoaded` allows the caller of the function to define the value of the parameter `waitMs`. I can do `waitUntilDataIsLoaded(0)`, `waitUntilDataIsLoaded(6000)`, and `waitUntilDataIsLoaded(-123)` all of them are valid and will work. The `500` is only the default value if the caller does not define a value for `waitMs`.

While I think you are right on having hardcoded things are not good, I would not call this a hardcoded value, but just default values that are applied if the called does not define parameters.

Note, all variables have default values, for instance the default value of a number could be `0`, and the default value of a boolean can be `false`. All the function `waitUntilDataIsLoaded` is doing is to set the default value, not imposing a value that cannot be changed

vobratil: @carlosthe19916 Oh, you're right, my mistake for reading it as an assigned parameter, not as a default value. In that case approved :).

PR #744, opened 2025-09-23
fix: use original filename for the csv downloadable file

carlosthe19916:
Feedback was given after a short internal demo.
The filename used for downloading the CSV file should reuse the original name of the file selected for the analysis (SBOM filename)

## Summary by Sourcery

Reuse the original uploaded SBOM filename(s) for the downloadable CSV file and related UI messages.

Enhancements:
- Compute a joined filename from the uploaded file(s) via React.useMemo
- Use the computed filename for the CSV download instead of a static name
- Update the EmptyState message to reference the computed filename


carlosthe19916:
@gildub do you mind reviewing this PR please?

gildub (approved):

Issue #750, opened 2025-09-28
After an SBOM is deleted, there are errors displayed in the Dashboard


carlosthe19916:
Steps to reproduce:
- In the Dashboard select an SBOM to be rendered in the Donut Charts
- Go to the SBOMs List page and delete the SBOM you selected to watch in the Dashboard
- Go back to the Dashboard page and you will see an error in the DonutChart where you selected the SBOM 

As the backend do not clear the userPreferences data, the UI can clear SBOMs deleted assuming that if while fetching SBOMs the response contains a 404 error code then it means the SBOM has been deleted

PR #751, opened 2025-09-28
fix: dashboard - handle deleted sboms

carlosthe19916:
fixes: https://github.com/guacsec/trustify-ui/issues/750

If the SBOM fetch contains 404 then the dashboard will assume the SBOM has been deleting and update the user preferences accordingly.

## Summary by Sourcery

Handle deleted SBOMs in the dashboard by automatically clearing user preferences on 404 fetch errors and updating query caches accordingly

Bug Fixes:
- Clear watched SBOM preference when fetching a SBOM returns a 404 error

Enhancements:
- Extend useFetchSBOMById hook to accept refetchInterval and retry options
- Include deletion and ongoing mutation states in the loading indicator
- Invalidate dashboard queries and remove individual SBOM cache after deletion
- Allow null values in watched SBOM context patch and memoize the patch function with useCallback

### DIFF
--- a/client/src/app/pages/home/components/WatchedSbom.tsx
+++ b/client/src/app/pages/home/components/WatchedSbom.tsx
@@ -1,6 +1,8 @@
 import React from "react";
 import { generatePath, Link } from "react-router-dom";
 
+import { useIsMutating } from "@tanstack/react-query";
+
 import {
   Button,
   Card,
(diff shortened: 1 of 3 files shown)


client/src/app/queries/sboms.ts
```diff
@@ -116,6 +122,10 @@ export const useDeleteSbomMutation = (
     onSuccess: async (response, id) => {
       onSuccess(response, id);
       await queryClient.invalidateQueries({ queryKey: [SBOMsQueryKey] });
+
+      // Required as SBOMs in other pages might need to to be removed
+      // https://github.com/TanStack/query/discussions/3169#discussioncomment-14347395
```
gildub, line 127: I don't think this comment is needed.
`removeQueries()` goes on par with `invalidateQueries()`, we don't want need to delve into React Queries doco even though it needs to be updated.
If you want to keep it then maybe "`removeQueries()` updates the cache" or equivalent.

client/src/app/pages/home/components/WatchedSbom.tsx
```diff
@@ -61,7 +63,23 @@ export const WatchedSbom: React.FC<WatchedSbomProps> = ({
     sbom: currentSbom,
     isFetching: isFetchingCurrentSbom,
     fetchError: fetchErrorCurrentSbom,
-  } = useFetchSBOMById(sbomId ?? undefined);
+  } = useFetchSBOMById(sbomId ?? undefined, undefined, 1);
+
+  // Handle SBOMs that no longer exist
```
gildub, line 68: Don't think that comment is needed.
But this is just me not being a big fan of comments.

gildub (approved):
Looks good ready to go, just a cosmetic remark about comments and the need to add another PR to follow-up.

PR #822, opened 2025-11-04
tests: make Toolbar typesafe

carlosthe19916:
This PR enhances the Toolbar definition so we define the available filters at the time of instantiating the Toolbar, then when we apply filters we can only use the filters defined at the beginning. We enhance type safety and use the power of Typescript.

The following is how a toolbar will be instantiated after this PR:

```ts
const toolbar = await Toolbar.build(this._page, "advisory-toolbar", {
      "Filter text": "string",
      Revision: "dateRange",
      Label: "typeahead",
});

await toolbar.applyFilter({ "Filter text": "something" }); // Correct
await toolbar.applyFilter({ "Filter textttt": "something" }); // Typescript error: "Filter textttt" was not defined in Toolbar
await toolbar.applyFilter({ "Filter textttt": 123 }); // Typescript error: ""Filter text" only accept strings

await toolbar.applyFilter({ "Revision": {from: "01/01/2024", to: "01/01/2024" }); // Correct
```

Note: the current e2e tests are failing due to a backend issue that does not allow uploading parallel files, the backend engineers are working on fixing it. When we review this PR we can focus on the code and it won't be merged until the e2e tests are healthy again

### DIFF
--- a/e2e/tests/ui/pages/sbom-details/vulnerabilities/donutchart.spec.ts
+++ b/e2e/tests/ui/pages/sbom-details/vulnerabilities/donutchart.spec.ts
@@ -16,8 +16,8 @@ test.describe("DonutChart validations", { tag: "@tier1" }, () => {
 
     await expect(page.locator("#legend-labels-0")).toContainText("Critical: 0");
     await expect(page.locator("#legend-labels-1")).toContainText("High: 2");
-    await expect(page.locator("#legend-labels-2")).toContainText("Medium: 14");
-    await expect(page.locator("#legend-labels-3")).toContainText("Low: 0");
+    await expect(page.locator("#legend-labels-2")).toContainText("Medium: 13");
+    await expect(page.locator("#legend-labels-3")).toContainText("Low: 1");
     await expect(page.locator("#legend-labels-4")).toContainText("None: 0");
     await expect(page.locator("#legend-labels-5")).toContainText("Unknown: 0");
   });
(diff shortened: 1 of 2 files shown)


e2e/tests/ui/pages/Toolbar.ts
```diff
@@ -1,81 +1,126 @@
...
+  return type === "multiSelect";
+}
+
+function isTypeaheadFilter<K extends string, T extends Record<K, TFilterValue>>(
+  type: T[K],
+  value: unknown,
+): value is TMultiValue {
+  return type === "typeahead";
+}
+
```
mrrajan: `value` on these functions were never used.
I see the purpose this code block is to return the type of the filtering condition. As we retrieve filtering type on line [81](https://github.com/guacsec/trustify-ui/pull/822/files#diff-b9bdf33fd65e9cb221148c89f51820bb675144a96588f2cfb79864ed4b5bfb6fR81), this can be simply used with if condition without these functions 
```
if (filterType === "string") {
  await this.applyTextFilter(filterName as keyof TFilter, filterValue as string);
}
```
instead of
```
if (isStringFilter(filterType, filterValue)) {
        await this.applyTextFilter(filterName, filterValue);
}
```

mrrajan (changes requested):
@carlosthe19916 While running the tests locally, there are couple of errors occurred for CVSS score and severity which is due to known issue. But the sorting validation on SBOM list page and package details page - `ui/pages/sbom-list/sort.spec.ts:13` and `ui/pages/sbom-details/packages/sort.spec.ts:13` failed. Even though these errors not related to these changes, just highlighting that the sorting has been changed on the UI. 

I am running against upstream release/0.4.z branch.

PR #831, opened 2025-11-21
tests: avoid hard coding the list of imported sboms/advisories

carlosthe19916:
## Remove hardcoded list of files to be ingested from constants.ts

e2e/tests/common/constants.ts (main branch) contains constant variables that define the list of SBOM/Advisory to be ingested.

**Problem**:
- Each time we need to add new SBOMs/Advisories we need to manually maintain e2e/tests/common/constants.ts and it is getting too big

**Proposal**:
- Remove the constants from e2e/tests/common/constants.ts
- `readDirectoryRecursively` is a function that reads recursively files inside a directory. So instead of listing the files we want to ingest we only need to define the directory we want to ingest.

Some minor changes alongside the changes described above:
- Instead of having 2 functions for uploading SBOMs and Advisories, there is only 1 function named `uploadFiles`

## Organize sbom and advisory files

The directory e2e/tests/common/assets/csaf is getting messy as all new files always end up in that directory and its difficult to differentiate those files and their purpose.

**Proposal**:

The following is the new tree structure proposed:

```
.
├── assets
│   └── sbom
└── dataset
    ├── advisory
    │   ├── csaf
    │   ├── csaf_security
    │   ├── cve
    │   └── osv
    └── sbom
```

- `assets` directory should contain files that are used by individual tests. E.g.

https://github.com/guacsec/trustify-ui/blob/35fe4f5b9162103a85011b5b619b6d5dafcd9a32/e2e/tests/ui/features/%40sbom-scan/scan-sbom.feature#L25-L26

- `dataset` should contain files that are ingested in the `setup` phase by the file https://github.com/guacsec/trustify-ui/blob/main/e2e/tests/api/dependencies/global.setup.ts

This change is only moving files and not touching tests, this way we guarantee all tests remain green

## Summary by Sourcery

Simplify e2e test data ingestion by dynamically discovering SBOM and advisory files instead of relying on hardcoded lists.

Enhancements:
- Replace separate SBOM and advisory upload helpers with a single generic uploadFiles helper that posts to the appropriate API endpoint.
- Introduce a recursive directory reader to locate all SBOM and advisory files under common dataset directories, improving maintainability of test datasets.

Tests:
- Adjust global e2e setup to ingest SBOM and advisory data from dataset directories discovered at runtime rather than from hardcoded file name lists in constants.


mrrajan (approved):
LGTM

PR #844, opened 2025-12-03
fix: SBOM Details - Packages Tab - Long licenses occupy too much space

stanislavsemeniuk:
… for versions column

## Summary by Sourcery

Enhancements:
- Change the licenses table column to use truncation for long content instead of breaking words, matching the versions column behavior.

### DIFF
--- a/client/src/app/pages/sbom-details/packages-by-sbom.tsx
+++ b/client/src/app/pages/sbom-details/packages-by-sbom.tsx
@@ -217,20 +217,15 @@ export const PackagesBySbom: React.FC<PackagesProps> = ({ sbomId }) => {
                     </Td>
                     <Td
                       width={20}
-                      modifier="truncate"
+                      modifier="breakWord"
                       {...getTdProps({
                         columnKey: "licenses",
-                        isCompoundExpandToggle: item.licenses.length > 1,
+                        isCompoundExpandToggle: item.licenses.length > 0,
                         item: item,
                         rowIndex,
                       })}
                     >
-                      {item.licenses.length === 1
-                        ? renderLicenseWithMappings(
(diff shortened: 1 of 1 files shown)


carlosthe19916 (changes requested):
@stanislavsemeniuk I figured out this PR is attempting to solve https://issues.redhat.com/browse/TC-3051 

- Link this PR to the JIRA ticket so we tack things
- Update the status of the jira ticket to "in-progress" or equivalent

Seems to me that the changes in this PR are not aligned with Marco's request in the Jira Comment:
```
 the most consistent way to manage this looks like having always the license number in the package row and display the actual license value always and only in the expanded row.
For the specific situation, it would mean:

the row in the packages table for the "kernel" page will have "1 License" in the "Licenses" column
when the user clicks on "1 License", the expanded row will have the long license test displayed in multiple longer lines
```

What we need to do is to always make sure the column has an expanded area. Right now the expanded area is conditional.

client/src/app/pages/sbom-details/packages-by-sbom.tsx
```diff
@@ -220,17 +220,12 @@ export const PackagesBySbom: React.FC<PackagesProps> = ({ sbomId }) => {
                       modifier="breakWord"
                       {...getTdProps({
                         columnKey: "licenses",
-                        isCompoundExpandToggle: item.licenses.length > 1,
+                        isCompoundExpandToggle: item.licenses.length > 0,
                         item: item,
                         rowIndex,
                       })}
```
carlosthe19916: There is no need to add the compound properties now. This block of code can become:

```ts
{...getTdProps({
  columnKey: "licenses",
})}
```

PR #855, opened 2025-12-19
feat: manage columns in package list page

carlosthe19916:
This PR is just a proposal, it needs to be validated with UX and Engineers:

The Package List table contains a column `Path` and from a personal experience that columns is most of the time empty. It occupies space.

This PR adds the ability to show/hide table columns based on the user preferences. The saved columns will survive even if the user leaves the page because we are using the browser storage.

We are also able to define defaults in regards which columns could be hidden from the beginning, the column `path` could potentially be hidden as part of the default configuration.

https://github.com/user-attachments/assets/e2fe3633-ab14-4f30-9564-9eb7879ffedf

## Summary by Sourcery

Add configurable column visibility to the package list table and expose a UI to manage visible columns.

New Features:
- Allow users to show or hide individual columns in the package list table, with visibility persisted via table column state.
- Add a toolbar column-management control using PatternFly's ColumnManagementModal to configure which package list columns are visible.

Enhancements:
- Define initial and default column metadata for the package list, ensuring identity columns remain visible and disabling unused table selection/active-item features.

Build:
- Add @patternfly/react-component-groups as a client dependency for column management UI components.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -22,6 +22,7 @@
     "@hookform/resolvers": "^2.9.11",
     "@patternfly/patternfly": "^6.4.0",
     "@patternfly/react-charts": "^8.4.0",
+    "@patternfly/react-component-groups": "^6.4.0",
     "@patternfly/react-core": "^6.4.0",
     "@patternfly/react-log-viewer": "^6.3.0",
     "@patternfly/react-table": "^6.4.0",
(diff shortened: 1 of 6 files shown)


carlosthe19916:
Closing this as this was a experiment. First, needs to go through UX

PR #859, opened 2025-12-23
fix: add not-found page & route

stanislavsemeniuk:
https://issues.redhat.com/browse/TC-3218

Added not found page component and implemented specific route. Maybe worth improving 404 not found page functionality so it would appear even if route exists but item with specific id is no longer exist. (f.e when advisory was deleted)

## Summary by Sourcery

Add a dedicated 404 Not Found page and integrate it into routing and data loading error handling for key detail views.

New Features:
- Introduce a reusable 404 Not Found page component and route for unmatched paths.

Enhancements:
- Add a route-level error boundary that shows appropriate UX for 404, 401, and other HTTP errors on advisory, package, SBOM, and vulnerability detail routes.
- Refactor advisory, package, SBOM, and vulnerability queries to expose shared query option builders for reuse by route loaders.
- Centralize React Query client configuration into a shared module and reuse it in the app entrypoint.

### DIFF
--- a/client/src/app/Routes.tsx
+++ b/client/src/app/Routes.tsx
@@ -105,15 +105,6 @@ export const AppRoutes = createBrowserRouter([
             component={<AdvisoryDetails />}
           />
         ),
-      },
-      {
-        path: Paths.advisoryUpload,
-        element: (
-          <LazyRouteElement
-            identifier="advisory-upload"
-            component={<AdvisoryUpload />}
-          />
-        ),
         errorElement: <RouteErrorBoundary />,
         loader: async ({ params }) => {
           const advisoryId = usePathFromParams(params, PathParam.ADVISORY_ID);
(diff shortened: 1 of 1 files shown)


carlosthe19916 (changes requested):
Image

- Could we evaluate the removal of the word "no longer"? `no longer` means the page used to exist but now it doesn't exist anymore. Could we just say `404: That page does not exists`?
  - If I navigate to `http://localhost:3000/something` then the message is `404: That page no longer exists` however the page `/something` never existed.
- Could we evaluate the wording `so try searching TPA.`?
  -  Searching where or what? it is not very clear.
  - `TPA` is the Red Hat downstream brand and can only be used in the context of Red Hat products. However, this repository mainly serves the community version called `Trustify`
  - `Another page might have the information you need` could be enough.

As a general pattern, we should try to avoid using the brand names in any page except specific places like the "About" Modal or specific places. E.g. Avoid using Trustify or TPA in messages, tooltips, pages, etc. because in the long term it generates differences between Upstream and Downstream versions.

PR #869, opened 2026-01-03
tests(fix): reuse toHaveColumnWithValue function 

carlosthe19916:
- Enhances `e2e/tests/ui/assertions/TableMatchers.ts` so the function `toHaveColumnWithValue` can also match an exact row index if it is defined.
- Apply the function `toHaveColumnWithValue` where possible and avoid manual selection of columns

## Summary by Sourcery

Unify table column assertions in e2e UI tests by enhancing the table matcher to support optional row-index matching and replacing direct cell locator checks with the shared matcher.

Enhancements:
- Extend the custom table matcher `toHaveColumnWithValue` to optionally assert against a specific row index in a column.
- Refactor multiple UI column validation specs to use the shared `toHaveColumnWithValue` matcher instead of manual table cell locators for consistency and readability.

Tests:
- Update existing UI E2E column validation tests across SBOM, package, advisory, and vulnerability pages to rely on the enhanced `toHaveColumnWithValue` matcher.


carlosthe19916:
@sourcery-ai review

PR #878, opened 2026-01-09
feat: add claude docs

carlosthe19916:
This PR adds markdown files specific to Claude AI structure.
- The root of the project contains a file CLAUDE.md which should describe all the project not just the e2e directory because potentially we could use CLAUDE.md not just for the e2e directory but for other directories as well.
- The `.claude` directory contains 2 agents and 1 command.

Claude AI requires specific file names so I am not sure how to decouple it so Cursor users can also find benefit maintaining the markdown files defined in this directory. Ideas are welcome.

As this PR only adds markdown files, it actually represents low risk even if its content is not 100% perfect.

## Summary by Sourcery

Add Claude-specific project documentation and agent/command configurations to support AI-assisted development workflows for the Trustify UI repository.

Documentation:
- Introduce CLAUDE.md with architecture overview, workspace structure, development workflows, and coding standards for the Trustify UI project.

Chores:
- Add Claude agent definitions for UX and e2e test review plus an E2E test generation command and related local configuration files.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,6 +33,22 @@ This monorepo uses 4 npm workspaces:
   - `tests/ui/` - UI tests with Page Object Model
   - `tests/ui/features/` - BDD tests with Gherkin + playwright-bdd
 
+## Environment Variables
+
+Defined in `common/src/environment.ts`:
+
+- `TRUSTIFY_API_URL` - Backend API URL (default: `http://localhost:8080`)
+- `AUTH_REQUIRED` - Enable/disable authentication (default: `"true"`)
+- `OIDC_SERVER_URL` - Keycloak/OIDC server URL
+- `OIDC_CLIENT_ID` - OIDC client ID (default: `"frontend"`)
+- `OIDC_SCOPE` - OIDC scope (default: `"openid"`)
+
+**How env vars work:**
+
+- **Production**: Server injects env as base64 JSON into `index.html.ejs`,
+  client decodes from `window._env`
+- **Development**: Rsbuild injects env directly into HTML template
+
 ## Essential Commands
 
 ### Development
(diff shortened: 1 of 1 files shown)


.claude/commands/e2e-test.md
```diff
@@ -0,0 +1,36 @@
...
+## Task
+
+Create E2E Playwright tests based on [summary] and following these guidelines:
+
+- Tests are written under `e2e/tests/ui`
+
+## Review the work
+
+- **Invoke the e2e-test-reviewer subagent** to review your work and implement
+  suggestions where needed/
```
mrrajan: Remove the slash at the end.

mrrajan (changes requested):
Thanks for the PR @carlosthe19916, I have added some review comments. 

I personally prefer that the sub agent to have clearly defined role, either as a reviewer or writer. Lets explore the possibilities after this PR.

carlosthe19916:
> Thanks for the PR @carlosthe19916, I have added some review comments.
> 
> I personally prefer that the sub agent to have clearly defined role, either as a reviewer or writer. Lets explore the possibilities after this PR.

Maybe the existence of a "command" might have created a confusion... we can iterate this whole thing in a future PR.

- A **command** is a direct instruction you give to Claude in the current conversation.
It tells Claude what to do right now, **in one step**.
  - **Example**: “Summarize this document in bullet points.”
  - Runs inline in the same context. Single-step or short multi-step reasoning. No memory or autonomy beyond the prompt
- A "sub agent" is a delegated instance of Claude created to handle a specific role or task independently. We're essentially saying: “Go do this as a separate worker, then come back with results.” Then if needed iterate. So it usually represents a **multi-step task**
  - Multiple sub agents can run in parallel 
  - Good if:
    - The task is complex or multi-stage
    - We want division of labor
    - Different perspectives or skills are helpful
    - We want Claude to think independently about part of the problem

The current PR is using the command mainly for invoking the sub agent. I thought that if we had multiple agents then we could use commands to invoke one or more sub agents in parallel so the command's role is to orchestrate sub agents. Not sure if that is what we want so I am open to enhancements.

Again, i would like little by title to enhance all of this and cover not just the testing part but also the development part (I will coordinate these efforts with the UI team just like we do for the QE part)

PR #886, opened 2026-01-22
Grouping for Product - Group create/edit front end

stanislavsemeniuk:

https://issues.redhat.com/browse/TC-3447

Implemented SBOMGroupFormModal component that can be used as a form for creating and editing sbom group. Used react hook form and patternfly input components. Also implemented reusable component for adding labels that can be used inside hook form - HookFormPFAddLabels and added MenuWithDrilldown from patternfly to our codebase.

Screenshot 2026-01-22 at 12 47 54

## Summary by Sourcery

Add a modal-driven form for creating and editing SBOM groups and wire it into the SBOM list toolbar.

New Features:
- Introduce SBOMGroupFormModal to capture SBOM group details including name, parent group, product flag, description, and labels.
- Add a reusable HookFormPFAddLabels field component for managing metadata labels within react-hook-form.
- Expose controllerProps in HookFormPFGroupController to allow passing validation and other Controller options from consumers.
- Add a Create Group action to the SBOM toolbar that opens the new group form modal.

## Summary by Sourcery

Add a modal form and supporting infrastructure to create SBOM groups from the SBOM list toolbar, including hierarchical group selection, label management, and backend integration.

New Features:
- Introduce SBOMGroupFormModal for creating and editing SBOM groups with fields for parent group, product flag, description, and metadata labels.
- Add HookFormPFAddLabels as a reusable react-hook-form field for managing key-value metadata labels with validation rules.
- Add a hierarchical group selector (HookFormPFGroupSelect) backed by a drilldown menu UI for browsing and searching nested SBOM groups.
- Enable SBOM group creation from the SBOM toolbar, wired to backend mutations and user notifications.

Enhancements:
- Expose controllerProps on HookFormPFGroupController to allow consumers to pass react-hook-form Controller options such as validation rules.
- Add a reusable MenuWithDrilldown component wrapping PatternFly's drilldown menu with search and selection handling.
- Introduce React Query hooks and helpers (useFetchSBOMGroups, useCreateSBOMGroupMutation, checkGroupNameUniqueness) to list and create SBOM groups and validate name uniqueness.


client/src/app/queries/sbom-groups.ts
```diff
@@ -0,0 +1,77 @@
...
+      },
+    });
+    const isUnique = !response.data?.items || response.data?.items.length === 0;
+    return isUnique;
+  } catch (error) {
+    // On error, assume name is available (fail open for better UX)
+    console.error("Failed to check group name uniqueness:", error);
+    return true;
+  }
+};
```
carlosthe19916: let's move this function to `api/model-utils.ts` since it is not used directly by any of the React Queries

client/src/app/api/model-utils.ts
```diff
@@ -182,3 +184,31 @@ export const extractPriorityScoreFromScores = (scores: Score[]) => {
 
   return [...scores].sort(compareByScoreTypeFn((item) => item.type))[0];
 };
+
```
CryptoRodeo: ✔️

client/src/app/pages/sbom-list/components/CreateGroupForm/SBOMGroupFormModal.tsx
```diff
@@ -0,0 +1,290 @@
...
+    defaultValues: {
+      ...defaultValues,
+      ...initialValues,
+    },
+  });
+
+  const parentGroup = watch("parentGroup");
+  const isProduct = watch("isProduct");
+
+  React.useEffect(() => {
```
CryptoRodeo: I wonder if we can sync this form in a non-imperative manner rather than use a useEffect hook 🤔 I _think_  the form stays mounted even when the modal closes here, so the state needs to be synced imperatively on reopen every time. And because useEffect + reset() runs after render, there may be some subtle UI issues like brief flashes of stale values...

Maybe something like this could help?

You can have an outer component here that doesn't mount the form if it's not open:
```typescript
export const SBOMGroupFormModal: React.FC<SBOMGroupFormModalProps> = (
  props,
) => {
  if (!props.isOpen) return null;
  return <SBOMGroupFormModalInner {...props} />;
};
```
And then an inner component with most of the code you already have in this component but now `useForm` initializes with the correct `defaultValues` on mount with no timing concerns from `useEffect` and `reset()`.

```typescript
const SBOMGroupFormModalInner: React.FC<SBOMGroupFormModalProps> = ({
  onClose,
  onSubmit,
  initialValues,
  type = "Create",
}) => {
  const [isAdvancedExpanded, setIsAdvancedExpanded] =
    React.useState<boolean>(false);
  const [tempName, setTempName] = React.useState<string>("");
  const [tempParent, setTempParent] = React.useState<Group | undefined>(
    undefined,
  );

  const {
    control,
    handleSubmit,
    getValues,
    watch,
    trigger,
    formState: { isSubmitting, isValid, isValidating },
  } = useForm<SBOMGroupFormValues>({
    mode: "onChange",
    defaultValues: {
      ...defaultValues,
      ...initialValues,
    },
  });
....
```
This is just a suggestion though, not a blocker.

CryptoRodeo (commented):
Hey, I will take a closer look at this on Monday if this is not merged by then (I'm almost done). In the mean time here is a comment on an idea. Not a blocker.

client/src/app/pages/sbom-list/sbom-toolbar.tsx
```diff
@@ -49,54 +80,70 @@ export const SbomToolbar: React.FC<SbomToolbarProps> = ({
     propHelpers: { toolbarBulkSelectorProps },
   } = bulkSelectionControls;
 
+  const handleCreateGroup = (body: GroupRequest) => {
+    createGroupMutation.mutate(body);
+  };
+
```
carlosthe19916: Minor thing but could we move this function to line 61? this is a function related to group and all the rest of code for groups are between line 38-60. Just to read is easier.

PR #909, opened 2026-02-02
chore(claude): add format and linting hook

carlosthe19916:
See https://code.claude.com/docs/en/hooks

- The current Hook will be executed every time a file is created or edited
  - When Claude code touches a file it generates a JSON similar to https://code.claude.com/docs/en/hooks#posttooluse-input that is later automatically injected into the Hook, this way we know which exact file we need to verify Linting and Format against
- The script `.claude/hooks/lint-and-format.sh` will run against the file created/edited to run Biome (Linting and Formatting) 
  - Biome is executed as `npx biome check --write "$FILE_PATH"` with the `--write` param to automatically fix fixable format and linting issues
- Thje hook is a non-blocking hook. We should not block the whole Claude Code workflow in case Biome (format or linting) fails and let AI to continue with its workflow

Applying hooks for deterministic tasks will help us to better automate the AI workflows we have at the moment

## Summary by Sourcery

Add a Claude hook to automatically format, lint, and type-check files touched by AI-generated changes.

Enhancements:
- Introduce a Claude hook script that runs Biome formatting/linting and TypeScript type-checking on modified files by workspace.

Build:
- Configure Claude project settings to enable the new lint-and-format hook.

### DIFF
--- a/.claude/hooks/lint-and-format.sh
+++ b/.claude/hooks/lint-and-format.sh
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# Check if jq is available, exit gracefully if not
+if ! command -v jq &> /dev/null; then
+  echo "⚠️  jq is not installed, skipping lint and format checks" >&2
+  exit 0
+fi
+
 # Read hook input from stdin
 INPUT=$(cat)
 


CryptoRodeo (approved):
Cool!

PR #918, opened 2026-02-05
tests: Add API tests for sorting by various categories on the Search page

matejnesuta:
Adding this, as it is a more convenient and less flaky approach of testing the sorting than directly in the UI.

## Summary by Sourcery

Add API-level tests to verify sorting behavior for PURLs, vulnerabilities, and SBOMs across multiple sortable fields.

Tests:
- Add PURL sorting tests covering name, namespace, and version in both ascending and descending orders.
- Add vulnerability sorting tests validating ID, CVSS score, and published date ordering for ascending and descending sort directions.
- Add SBOM sorting tests that assert correct name and published-date ordering for ascending and descending sorts.

### DIFF
--- a/e2e/tests/api/helpers/sorting-helpers.ts
+++ b/e2e/tests/api/helpers/sorting-helpers.ts
@@ -2,22 +2,25 @@ import type { AxiosInstance } from "axios";
 import { expect } from "../fixtures";
 
 /**
- * Helper to validate that dates in an array are sorted in the specified order
+ * Generic helper to validate sorting with optional value transformation
  */
-export function validateDateSorting(
+function validateSorting(
   // biome-ignore lint/suspicious/noExplicitAny: Generic helper accepts any API response type
   items: any[],
-  dateField: string,
+  field: string,
   order: "ascending" | "descending",
+  // biome-ignore lint/suspicious/noExplicitAny: Transform function accepts any value type
+  transform?: (value: any) => any,
 ) {
-  // Extract date values and convert to timestamps for comparison
+  // Extract values and optionally transform them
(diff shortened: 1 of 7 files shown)


e2e/tests/api/helpers/sorting-helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+    return b - a;
+  });
+
+  expect(dates).toEqual(sorted);
+}
+
+/**
+ * Helper to validate that numeric scores in an array are sorted in the specified order
+ */
+export function validateNumericSorting(
```
vobratil: Again, just a nitpick, but this method and `validateDateSorting` could probably share a big part of the body, since the only difference seems to be in the sorted data type.

e2e/tests/api/helpers/sorting-helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+export async function validateSortDirectionDiffers(
+  axios: AxiosInstance,
+  endpoint: string,
+  sortField: string,
+  // biome-ignore lint/suspicious/noExplicitAny: Generic helper accepts any API response type
+  extractValue: (item: any) => any,
+) {
+  const [ascResponse, descResponse] = await Promise.all([
+    axios.get(endpoint, {
+      params: { offset: 0, limit: 100, sort: `${sortField}:asc` },
```
vobratil: Oh, this is neat, I should start writing my requests like this.

vobratil (changes requested):
I would say that this PR looks pretty good, no major changes needed. The only thing I would like to ask for is moving the helpers file, so that we keep at least some order in the repo. And then there are my questions about why we are okay with the basic sort in some places and actually do a proper validation elsewhere.

PR #934, opened 2026-02-25
tests: Add Claude agents for API testing

matejnesuta:
## Summary by Sourcery

Add Claude agent configurations to support automated generation, review, orchestration, and coverage analysis for Playwright-based API tests in Trustify UI.

New Features:
- Introduce an API test reviewer agent configuration to enforce project standards, run linting, and score Playwright API tests.
- Introduce an API test generator agent configuration to create Playwright API tests from the OpenAPI spec while respecting existing test structure.
- Introduce an API test orchestrator agent configuration to coordinate iterative test generation and review workflows with quality gates.
- Introduce an API coverage analyzer agent configuration to compare OpenAPI definitions against existing tests and highlight coverage gaps.

### DIFF
--- a/.claude/agents/api-test-generator.md
+++ b/.claude/agents/api-test-generator.md
@@ -1,0 +1,698 @@
+---
+name: api-test-generator
+description: |
+  Generate Playwright API tests for Trustify endpoints.
+
+  Reusable agent that generates tests for single or multiple endpoints based on
+  OpenAPI spec. Follows existing test patterns and can accept feedback for iteration.
+
+  Works with api-test-orchestrator for bulk generation with quality checks.
+model: sonnet
+---
+
+You are the API Test Generator for Trustify UI. You generate Playwright API integration tests based on the OpenAPI specification, following established project patterns.
+
+## Your Responsibilities
+
+1. **Parse OpenAPI spec** for endpoint details
+2. **Generate test code** following project patterns
+3. **Use existing datasets** when appropriate
+4. **Run tests** to verify functionality
+5. **Report results** with clear status
+6. **Accept feedback** from reviewer for iteration
+
+**IMPORTANT**: You ONLY generate tests. You do NOT review code quality - that's the reviewer's job.
+
+**CRITICAL**: When adding to existing files, ONLY add new tests. DO NOT refactor, reorganize, or "improve" existing code unless explicitly asked.
+
+## Core Patterns
+
+### File Structure
+
+**Location**: `e2e/tests/api/features/[domain].ts`
+
+**Template**:
+```typescript
+import { expect, test } from "../fixtures";
+
+// Test cases below...
+```
+
+### Test Patterns
+
+**Basic GET request**:
+```typescript
+test("Description of test", async ({ axios }) => {
+  const response = await axios.get("/api/v2/endpoint");
+
+  expect(response.status).toBe(200);
+  expect(response.data).toEqual(
+    expect.objectContaining({
+      field: expectedValue,
+    }),
+  );
+});
+```
+
+**GET with query parameters** (CRITICAL - Always use URLSearchParams):
+```typescript
+test("Filter with complex query", async ({ axios }) => {
+  const queryParams = new URLSearchParams();
+  queryParams.append("offset", "0");
+  queryParams.append("limit", "10");
+  queryParams.append("sort", "published:asc");
+  queryParams.append("q", "CVE-2023-2&average_severity=medium|high");
+
+  const response = await axios.get("/api/v2/endpoint", {
+    params: queryParams,
+  });
+
+  expect(response.status).toBe(200);
+  expect(response.data.total).toBe(expectedCount);
+});
+```
+
+**POST request**:
+```typescript
+test("Create resource", async ({ axios }) => {
+  const body = {
+    field1: "value1",
+    field2: "value2",
+  };
+
+  const response = await axios.post("/api/v2/endpoint", body);
+
+  expect(response.status).toBe(201);
+  expect(response.data).toEqual(
+    expect.objectContaining({
+      id: expect.any(String),
+    }),
+  );
+});
+```
+
+**Path parameters** (encode when needed):
+```typescript
+test("Get by ID", async ({ axios }) => {
+  const id = "some-id";
+  const encodedId = encodeURIComponent(id);
+
+  const response = await axios.get(`/api/v2/endpoint/${encodedId}`);
+
+  expect(response.status).toBe(200);
+});
+```
+
+**Negative testing**:
+```typescript
+test("Rejects invalid input", async ({ axios }) => {
+  const response = await axios
+    .post("/api/v2/endpoint", { invalid: "data" })
+    .catch((err) => err.response);
+
+  expect(response.status).toBe(400);
+});
+```
+
+**Test grouping**:
+```typescript
+test.describe("Feature Name - Test Category", () => {
+  const commonData = { ... };
+
+  test("test case 1", async ({ axios }) => { ... });
+  test("test case 2", async ({ axios }) => { ... });
+});
+```
+
+### Code Quality Standards
+
+1. **TypeScript**: Proper types, no `any`
+2. **Async/await**: All axios calls
+3. **Error handling**: Use `.catch((err) => err.response)` for negative tests
+4. **Clear test names**: Describe what is being tested
+5. **Assertions**: Use `objectContaining`, `arrayContaining` for partial matches
+6. **No hard-coded waits**: Tests should be deterministic
+
+## Generation Workflow
+
+### Step 1: Parse OpenAPI Specification
+
+**Input**: Endpoint path and method (e.g., "GET /api/v2/advisory")
+
+**Read**: `client/openapi/trustd.yaml`
+
+**Extract**:
+- operationId
+- Parameters (query, path, body)
+- Request body schema
+- Response schema (200, 400, 404, etc.)
+- Deprecated flag
+
+**Example**:
+```
+GET /api/v2/advisory:
+  operationId: listAdvisories
+  parameters:
+    - q (query, optional, string): Query DSL
+    - sort (query, optional, string): Sort fields
+    - limit (query, optional, integer): Max results
+    - offset (query, optional, integer): Skip results
+  responses:
+    200: PaginatedAdvisoryList
+      schema:
+        total: integer
+        items: array[AdvisoryHead]
+```
+
+### Step 2: Check for Existing Tests
+
+**Scan**: `e2e/tests/api/features/` directory
+
+**Check if endpoint already tested**:
+- Search for endpoint path in existing test files
+- Identify what's already covered (happy path, params, negative)
+- **If file exists, READ it to understand structure**
+
+**CRITICAL - Respecting Existing Code**:
+
+**If test file exists**:
+1. **READ the entire file first**
+2. **Identify existing structure**:
+   - Are tests flat or grouped in `describe` blocks?
+   - What naming conventions are used?
+   - What assertion patterns are used?
+3. **Match the existing style**:
+   - If tests are flat, add flat tests
+   - If tests use `describe`, add to existing `describe` or create new one
+   - Follow same naming pattern
+4. **ONLY append new tests** - DO NOT:
+   - Refactor existing tests
+   - Reorganize file structure
+   - Rename existing tests
+   - Change existing assertions
+   - Add `describe` blocks if file doesn't use them
+   - Remove `describe` blocks if file uses them
+
+**Example - File with flat tests**:
+```typescript
+// Existing file: purl.ts
+import { expect, test } from "../fixtures";
+
+test("Purl by alias - vanilla", async ({ axios }) => {
+  // existing test
+});
+
+// ADD NEW TEST HERE (flat, matching style):
+test("Purl by ID", async ({ axios }) => {
+  // new test
+});
+
+// DON'T DO THIS (adding describe when file is flat):
+test.describe("PURL Tests", () => {  // ❌ WRONG
+  test("Purl by ID", async ({ axios }) => { ... });
+});
+```
+
+**Example - File with describe blocks**:
+```typescript
+// Existing file: recommendation.ts
+import { expect, test } from "../fixtures";
+
+test.describe("Recommendation API - Invalid PURL Format", () => {
+  // existing tests
+});
+
+// ADD NEW DESCRIBE BLOCK HERE (matching style):
+test.describe("Recommendation API - Empty Results", () => {
+  test("Returns empty for unknown package", async ({ axios }) => {
+    // new test
+  });
+});
+```
+
+### Step 3: Identify Reusable Datasets
+
+**For file upload endpoints** (POST with multipart/form-data):
+
+**Scan** for existing dataset files:
+- SBOMs: `*.json`, `*.xml` (SPDX, CycloneDX)
+- CSAFs: advisory documents
+- VEXes: vulnerability exchange
+
+**Use existing files** - DO NOT create new ones:
+```typescript
+test("Upload SBOM", async ({ axios }) => {
+  // Use existing file: e2e/tests/api/data/quarkus-sbom.json
+  const fs = require('fs');
+  const filePath = path.join(__dirname, '../data/quarkus-sbom.json');
+  const fileContent = fs.readFileSync(filePath);
+
+  const formData = new FormData();
+  formData.append('file', fileContent, 'quarkus-sbom.json');
+
+  const response = await axios.post('/api/v2/sbom', formData, {
+    headers: formData.getHeaders(),
+  });
+
+  expect(response.status).toBe(201);
+});
+```
+
+**IMPORTANT**: Focus on API contract testing, NOT data format variations.
+
+### Step 4: Generate Test Code
+
+**Determine test type** (from user input or auto-detect):
+- Happy path (default)
+- Parameter variation
+- Negative testing
+- Complex operation
+- **Bugfix/regression test** (requires Jira ID)
+
+**Generate appropriate test**:
+
+**Happy path example**:
+```typescript
+test("List advisories", async ({ axios }) => {
+  const response = await axios.get("/api/v2/advisory?limit=10&offset=0");
+
+  expect(response.status).toBe(200);
+  expect(response.data).toEqual(
+    expect.objectContaining({
+      total: expect.any(Number),
+      items: expect.any(Array),
+    }),
+  );
+});
+```
+
+**Parameter variation example**:
+```typescript
+test("List advisories with complex query", async ({ axios }) => {
+  const queryParams = new URLSearchParams();
+  queryParams.append("offset", "0");
+  queryParams.append("limit", "10");
+  queryParams.append("sort", "modified:desc");
+  queryParams.append("q", "title=RHSA&average_severity=critical");
+
+  const response = await axios.get("/api/v2/advisory", {
+    params: queryParams,
+  });
+
+  expect(response.status).toBe(200);
+  expect(response.data.items.length).toBeLessThanOrEqual(10);
+});
+```
+
+**Negative test example**:
+```typescript
+test("Rejects invalid query syntax", async ({ axios }) => {
+  const response = await axios
+    .get("/api/v2/advisory?q=invalid_field=value")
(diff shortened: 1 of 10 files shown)


.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+
+```typescript
+import { expect, test } from "../fixtures";
+import { uploadFiles, deleteSboms } from "../helpers/general-helpers";
+```
+
+### Rules
+
+- **NEVER** import `test` or `expect` from `@playwright/test` directly — always use `"../fixtures"`
+- Omit `expect` from the import if only `test` is needed
```
vobratil, line 48: Maybe this should be a more general rule, something like "check if all imports are actually used in the code"?

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+## 3. Query Parameter Handling (CRITICAL)
+
+Choose the right approach based on query complexity:
+
+### When to Use `URLSearchParams`
+
+Use `URLSearchParams` when the query contains:
+- Special characters (`&`, `=`, `|`, spaces, etc.)
+- Complex query DSL via the `q` parameter
+- User input or dynamic variable values
```
vobratil: Maybe re-phrase this so that it's clear what exactly should be done with those?

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+- **Helpers**: `e2e/tests/api/helpers/[name].ts`
+- **File extension**: `.ts` (not `.spec.ts`)
+
+---
+
+## 3. Query Parameter Handling (CRITICAL)
+
+Choose the right approach based on query complexity:
+
+### When to Use `URLSearchParams`
```
vobratil: Why don't we actually enforce the usage of `URLSearchParams` always? Is there a downside, since we are not expecting to actually write the query? Plus `URLSearchParams` automatically URL encodes parameters, which would leave less space for mistakes when using purls of CPEs. I'm not heavily opinionated on this, just wondering why we would want several approaches.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+});
+```
+
+### When Inline URL Is Acceptable
+
+Simple static queries without any dynamic values MAY use an inline URL:
+
+```typescript
+// ✓ Acceptable for simple static-only queries
+const response = await axios.get("/api/v3/sbom?limit=10&offset=0");
```
vobratil: In addition to the URL formatting part, even though it may sound silly, I'm just wondering if it would make sense to keep all the .md files intended for AI to a minimum necessary length? The smaller the context, the smaller the token usage, I would assume, which could really add up on unnecessary parts over thousands of prompts. Maybe it's a small optimization, I don't know, but I think eventually AI pricing is going to become a concern.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+      id: expect.any(String),
+      identifier: expect.any(String),
+    }),
+  ]),
+);
+```
+
+### Deep Object Matching
+
+Use `toMatchObject` for deeply nested structures where the exact shape is important:
```
vobratil: While this is good advice for a human, I'm not sure if the agent has much visibility into the test data / is capable of deciding when the shape is important. In my experience AI really struggles with this and more often than not this has to be done manually. Just something to consider - do we need this part for AI, or should we just ask it to use placeholders instead?

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+// ❌ Too weak — asserts existence but not shape
+expect(response.data).toBeDefined();
+expect(response.data).not.toBeNull();
+```
+
+---
+
+## 5. Error Handling for Negative Tests (HIGH)
+
+Axios throws on non-2xx responses. Use `.catch((err) => err.response)` to capture error responses for assertion:
```
vobratil, line 213: I guess `throws on` may be a typo?

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+});
+```
+
+### ✅ ALSO ACCEPTABLE — `try/catch`
+
+```typescript
+test("Rejects invalid input", async ({ axios }) => {
+  try {
+    await axios.post("/api/v3/endpoint", { invalid: "data" });
+    throw new Error("Should have thrown");
```
vobratil, line 233: Is that a good example, actually? Shouldn't the `try` block be attempting its functionality "in good faith"? The throw should be unnecessary, as the backend should throw an error on its own.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+  } catch (error: unknown) {
+    const axiosError = error as { response: { status: number } };
+    expect(axiosError.response.status).toBe(400);
```
vobratil, line 236: This just leads me to the thought - when developing a test, should the AI also attempt to consider other than expected outcomes and throw descriptive logs about why the test went wrong? It's always so much better if the test tells you what went wrong than when you have to read the stack trace.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+});
+```
+
+---
+
+## 7. Code Reusability (HIGH)
+
+### Use Existing Helpers
+
+Before writing logic inline, check the existing helper files:
```
vobratil: It's a good point, but I don't know if we should list the explicit list of functions, as that is sure to change over time and we shouldn't have to update the list here as well.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+});
+```
+
+### Flag Duplication
+
+If new tests contain helper functions or repeated logic that already exists (or could be extracted), flag it:
+
+- Minor duplication (2–3 lines): **MEDIUM**
+- Helper function duplicated across files: **HIGH**
+- Duplication across many files: **CRITICAL** (extract to `e2e/tests/api/helpers/`)
```
vobratil, line 333: I don't really see much difference between high and critical. If logic from the helper functions is being duplicated, it should be always replaced with the corresponding function.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+
+### File Upload Tests
+
+Use `uploadFiles` from `general-helpers.ts` — never read files manually:
```
vobratil, line 337: Theoretically the explicit use of `uploadFiles` should be saved only for tests when we, for whatever reason, don't want the files to be uploaded during the initial document ingestion, like when testing the latest endpoint, where the presence of certain SBOMs could make the test pointless.

However, this leads me to a semi-related thought that we currently have no proper cleanup in our tests, except for a handful of my tests, where it was necessary, otherwise the tests would break after the first run. Perhaps that's something we could mention here? E.g. that the env should be brought back to the original state at the end of each test run.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+
+- Clear, descriptive test names: describe the domain, action, and expected outcome
+- Descriptive variable names: `response`, `items`, `queryParams` — not `r`, `d`, `qp`
+
+### Async/Await
+
+- Always `await` axios calls
+- No floating promises
+
+### No Hard-Coded Waits
```
vobratil, line 383: There's actually a few cases where this may be valid, e.g. in the case of Atlas tests, which uses specific timeouts. Maybe we could just say that timeouts should be parameterizable in the test header?

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+
+test.describe("SBOM upload", () => {
+  let sbomIds: string[] = [];
+
+  test.afterEach(async ({ axios }) => {
+    await deleteSboms(axios, sbomIds);
+    sbomIds = [];
+  });
+
+  test("Upload SBOM", async ({ axios }) => {
```
vobratil, line 350: Additionally, anything using file uploads and deletes should probably be run sequentially.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+
+test("Test B", async ({ axios }) => {
+  // All setup done inline, no dependency on Test A
+});
+```
+
+### ❌ BAD — Shared mutable state
+
+```typescript
+let sharedId: string;  // ❌ Module-level mutable state
```
vobratil, line 446: If we are giving `let` as an example here, maybe we should use `const` in the correct example.

.claude/shared/api-test-standards.md
```diff
@@ -0,0 +1,548 @@
...
+- Missing assertions on `response.data` structure (only checking status code)
+- Weak assertions (`toBeDefined`, `not.toBeNull`) when schema validation is possible
+- Helper functions duplicated across feature files
+- Modifying existing tests when asked to add new ones
+
+### MEDIUM (Nice to fix)
+
+- `URLSearchParams` used where a plain object or inline URL would be clearer
+- `any` type without a suppression comment explaining why
+- Unclear variable names (`r`, `d`)
```
vobratil, line 500: I think that in today's day and age this shouldn't appear in production-level code. I would put this to high.

vobratil (commented):
@matejnesuta This is a pretty hefty PR, so I've only reviewed the `api-test-standards` section. I will try to continue tomorrow, but hopefully you can take a look at the comments already. I've never reviewed a PR of this type, so I'm not sure if my comments are useful. Any feedback would be welcome :)

.claude/shared/api-test-standards.md
matejnesuta: "we currently have no proper cleanup in our tests"

I think this is kind of tied to a question, whether we need one in our tests or not. Back when I was in Service Mesh, we did not have any clean-ups in our tests. Instead, there was a some sort of "Before" condition, which prepared the environment for the test execution. This was done to make finding root cause of bugs easier in case some of the tests failed in our Jenkins pipelines. 

Maybe this is something to be discussed with the whole team.

.claude/shared/api-test-standards.md
matejnesuta: So, I understand that concern and I'll try to optimize the files once I address other issues in this PR. However, I am not really sure if it is going to change much in the grander scheme of things. I don't think we currently have any way of usage tracking of Claude across the whole Gaucsec. I have the same concern about the pricing, I just hope I won't spend way too much time on something, which did not need any optimalizations in the first place. 🥲

.claude/shared/api-test-standards.md
matejnesuta: I wonder if our linter already does this. It should be definitely possible without an LLM.

.claude/shared/api-test-standards.md
matejnesuta: Do you maybe have an example for this? I could not find anything in the code-base myself

.claude/shared/api-test-standards.md
matejnesuta: You are right, there probably isn't. I changed it to `URLSearchParams` only.

.claude/shared/api-test-standards.md
matejnesuta: I think it does work, but it is definitely uglier than the first variant. I'll change it.

.claude/shared/api-test-standards.md
matejnesuta: It seems like there are several ways of achieving this without having Claude hard coding it into every test, but I am not sure what should be the best approach for our use-case.

.claude/shared/api-test-standards.md
vobratil: True, now that you mention it, I'm pretty sure that it does. Disregard my comment then :).

.claude/shared/api-test-standards.md
vobratil: I think that even the Claude best practices document recommends being efficient with the context that the LLM uses, because it's not infinite and it fills up quickly. Anyway, just something to keep in mind. We don't need to meticulously go over each line and optimize, but it's probably a good idea to be economical with the prompts.

.claude/shared/api-test-standards.md
vobratil: Good point. Maybe we can bring it up at the nearest meeting.

.claude/shared/api-test-standards.md
vobratil: At the moment, the only test that I know of that does this is [one that I added](https://github.com/guacsec/trustify-ui/blob/681d588052f33ca23c4ce2936741ff311fb7dc39/e2e/tests/api/features/analysis-latest-atlas.ts#L32) recently. The fixed timeout is based on what the Atlas client CLI uses.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+You are the API Test Generator for Trustify UI. You generate Playwright API integration tests based on the OpenAPI specification, following established project patterns.
+
+## Standards Reference
+
+All code quality and style rules are defined in the shared standards document.
+Read and follow it before generating any tests:
+
+**[API Test Standards](../shared/api-test-standards.md)**
+
+Key sections:
```
vobratil, line 22: We had a discussion about this with Carlos and he claims that either the LLM ignores these references or includes them in the context at the end, which causes them not to work properly. Maybe it would be worth digging a little bit into this and trying to find out if this makes any difference?

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+- [Bugfix Tests](../shared/api-test-standards.md#9-bugfix--regression-tests-medium) — Jira ID format
+- [Test Independence](../shared/api-test-standards.md#10-test-independence-medium) — no shared mutable state
+- [Quick Reference Checklist](../shared/api-test-standards.md#12-quick-reference-checklist)
+
+## Your Responsibilities
+
+1. **Parse OpenAPI spec** for endpoint details
+2. **Generate test code** following the standards above
+3. **Reuse existing helpers** when appropriate
+4. **Run tests** to verify functionality
```
vobratil, line 40: This seems like a whole can of worms. I'm gonna leave a generic comment here and edit later as I go through the docs.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+5. **Report results** with clear status
+6. **Accept feedback** from reviewer for iteration
+
+**IMPORTANT**: You ONLY generate tests. You do NOT review code quality - that's the reviewer's job.
+
+**CRITICAL**: When adding to existing files, ONLY add new tests. DO NOT refactor, reorganize, or "improve" existing code unless explicitly asked.
```
vobratil, line 46: Appreciate the sarcasm towards the LLM 🤭

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+1. **READ the entire file first**
+2. **Identify existing structure**:
+   - Are tests flat or grouped in `describe` blocks?
+   - What naming conventions are used?
+   - What assertion patterns are used?
+3. **Match the existing style**:
+   - If tests are flat, add flat tests
+   - If tests use `describe`, add to existing `describe` or create new one
+   - Follow same naming pattern
+4. **ONLY append new tests** - DO NOT:
```
vobratil, line 100: This seems redundant with the part where you described the role of this agent.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+
+// ADD NEW DESCRIBE BLOCK HERE (matching style):
+test.describe("Recommendation API - Empty Results", () => {
+  test("Returns empty for unknown package", async ({ axios }) => {
+    // new test
+  });
+});
+```
+
+### Step 3: Identify Reusable Datasets
```
vobratil: This part is, in my opinion, just as good as telling the LLM to use a placeholder. I don't think that it can realistically read the whole dataset and decide which documents would be suitable for a given test case, as the context would be too large. If we are talking about being economical with the context usage, this would imho be a candidate for removal.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+
+**For file upload endpoints** (POST with multipart/form-data):
+
+**Scan** for existing dataset files:
+- SBOMs: `*.json`, `*.xml` (SPDX, CycloneDX)
+- CSAFs: advisory documents
+- VEXes: vulnerability exchange
+
+**Use existing files** - DO NOT create new ones:
```
vobratil, line 154: Now that I think about it, generating SBOMs using an LLM and some kind of parser that makes sure that the result is correct might be quite interesting. Not sure how practical that is.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+```typescript
+// Use Edit tool with old_string = last few lines of file
+// new_string = last few lines + new test
+```
+
+### Step 6: Run Test
+
+**Execute test**:
+```bash
+cd e2e
```
vobratil, line 264: Does the agent need to go back to the previous folder after test execution or does it always execute commands from a default location?

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+npm run test -- tests/api/features/[domain].ts
+```
+
+**Or run specific test**:
+```bash
+cd e2e
+npm run test -- tests/api/features/[domain].ts -g "test name"
```
vobratil, line 271: It should be specified in which cases to use which option. E.g. `if a test was added to an existing file, run only the newly created test`.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+    items: expect.any(Array),
+  }),
+);
+```
+
+### Step 3: Re-run Test
+
+After applying fixes:
+```bash
+cd e2e
```
vobratil, line 400: Same question as above.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+npm run test -- tests/api/features/[domain].ts
+```
+
+### Step 4: Report Fix Results
```
vobratil, line 404: I'm wondering if it wouldn't be actually simpler to have the results in a format that is a good compromise between human readibility and machine readibility, e.g. JSON. That would make them more reusable in case we want to actually do something with them.

.claude/agents/api-test-generator.md
```diff
@@ -0,0 +1,610 @@
...
+**Action**: Prompt for Jira ID if not provided
+**Output**: Test case with Jira comment and bug description
+
+### Mode 5: With Feedback (Iteration)
+**Input**: Feedback from reviewer
+**Output**: Fixed test file (only new tests modified)
+
+## Tools You'll Use
+
+- **Read**: OpenAPI spec, existing test files (ALWAYS read before editing), dataset files
```
vobratil: Again, I think that the dataset files are actually way too much for the LLM to take into consideration, literally hundreds of MB of text.

vobratil (commented):
Okay, this is unfortunately going much slower than I imagined, but I made it through the `api-test-generator.md` file. Left a few notes, nothing critical, but I have my doubts about the LLM's ability to work with the dataset in a meaningful way.

.claude/agents/api-test-generator.md
matejnesuta: I used a synthetic CSAF file for the Remediation feature testing and I think it is worth doing in cases, when it is too hard to find test files manually. I just did not want Claude to generate a new test file each time new tests are created.

.claude/agents/api-test-generator.md
matejnesuta: If I am not mistaken, shell state is not persistent across Bash tool calls.

.claude/agents/api-test-generator.md
matejnesuta: I have rewritten this part to prevent Claude from automatically scanning every test data file.

.claude/agents/api-test-generator.md
matejnesuta: I have rewritten this. Claude should now ask the user to provide test data.

.claude/agents/api-test-generator.md
matejnesuta: Addressed.

PR #944, opened 2026-03-04
tests: Adjustments to tests, which frequently fail on downstream instances

matejnesuta:
## Summary by Sourcery

Update flaky end-to-end tests for advisory and vulnerability explorers to match the current UI and reduce brittleness.

Tests:
- Adjust advisory explorer test to target the advisory ID via a more specific table cell locator.
- Relax vulnerability explorer scenario by removing a strict assertion on SBOMs table total results.


e2e/tests/ui/features/@vulnerability-explorer/vulnerability-explorer.feature
```diff
@@ -37,7 +37,6 @@ Feature: Vulnerability Explorer - View Vulnerability details
         When Tab "Related SBOMs" is visible
         Then The page title is "<vulnerabilityID>"
         Then The SBOMs table is sorted by "Name"
-        Then The SBOMs table total results is 1
```
mrrajan: I feel this step should work unless there is a additional SBOM affected by the vulnerability. I would suggest to verify if there are any additional sbom related to vulnerability and make use of table assertion method `toHaveNumberOfRows` to verify the expected row counts on the table.

mrrajan (changes requested):
@matejnesuta Thanks for the PR, Added suggestion below:

e2e/tests/ui/features/@vulnerability-explorer/vulnerability-explorer.feature
matejnesuta: Hello. I removed the line, because when importers are enabled, there is usually a big chance that some additional Quarkus SBOM gets loaded into the instance, which causes the test to fail on this line. The very next line already assumes that the correct SBOM should exist and I felt like this line only adds additional constraints without much benefit.

vobratil (approved):
This is good to go, let's not keep this open anymore.

mrrajan (approved):
LGTM

PR #989, opened 2026-04-14
Feat: CVSS Scores - Vulnerabilities Details page

stanislavsemeniuk:

https://redhat.atlassian.net/browse/TC-4092

Screenshot 2026-04-14 at 13 32 11

Also changed some e2e tests according to new logic

## Summary by Sourcery

New Features:
- Display a CVSS Assessment section on the vulnerability details page with scores grouped by CVSS version and highlighting the current base score type.


client/src/app/pages/vulnerability-details/vulnerability-details.tsx
```diff
@@ -84,6 +92,72 @@ export const VulnerabilityDetails: React.FC = () => {
...
+      {vulnerability &&
+        (() => {
+          const scoresByType = new Map<string, ScoredVector>();
+          for (const advisory of vulnerability.advisories) {
+            for (const score of advisory.scores) {
+              if (!scoresByType.has(score.type)) {
+                scoresByType.set(score.type, score);
+              }
+            }
+          }
```
carlosthe19916, line 104: The 3 Cards described in the mockups should entirely come from the vulnerability and not from the advisory. Or at least we need to make sure the Advisory belongs to the CVE project.

The mockups mentions: `Source: CVE Project`

I think there is a gap on the API or the mockups, because after looking at it manually then I am not sure which data is the right one to be used. I think we need to ask backend engs and UX

Image

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
```diff
@@ -84,6 +92,72 @@ export const VulnerabilityDetails: React.FC = () => {
...
+                      <CardFooter>
+                        <Content component="small">{score.vector}</Content>
+                      </CardFooter>
+                    </Card>
+                  ))}
+                </Gallery>
+              </PageSection>
+            </>
+          );
+        })()}
```
carlosthe19916, line 160: Let's avoid generating `()` a function.
I saw a lot it when I generated code with AI.

carlosthe19916 (commented):
I think I either discovered a gap on the rest api  vs mockups or I am missing something, because I am not sure which exact data should be populated in the cards. We need to ask UX and backend engs

carlosthe19916:
Just adding here the link of the issue https://github.com/guacsec/trustify/issues/2328 where it was concluded that the backend needs to change a bit to support this PR.

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
```diff
@@ -84,6 +93,53 @@ export const VulnerabilityDetails: React.FC = () => {
...
+      {vulnerability?.scores && (
+        <>
+          <Divider />
+          <PageSection>
+            <Content className={spacing.mbMd}>
+              <Content component="h2">CVSS Assessment</Content>
+              <Content component="small">Source: CVE project</Content>
+            </Content>
+            <Gallery hasGutter minWidths={{ default: "400px" }}>
+              {vulnerability?.scores.map((score) => (
```
kahboom, line 105: This might trigger a lint warning. We should at least provide the index here when mapping over arrays.

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
```diff
@@ -84,6 +93,53 @@ export const VulnerabilityDetails: React.FC = () => {
           </LoadingWrapper>
         </div>
       </PageSection>
+      {vulnerability?.scores && (
```
kahboom: This will render even if `vulnerability = []`–is that what we want? Maybe we could show an empty state for those and then have this here instead:
```
{vulnerability?.scores && vulnerability.scores.length > 0 && (
```

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
```diff
@@ -84,6 +93,53 @@ export const VulnerabilityDetails: React.FC = () => {
...
+        <>
+          <Divider />
+          <PageSection>
+            <Content className={spacing.mbMd}>
+              <Content component="h2">CVSS Assessment</Content>
+              <Content component="small">Source: CVE project</Content>
```
kahboom, line 102: Just wanted to bring it to your attention that this is still hard-coded. Will we be switching it to be dynamic? Does it need to be? Just checking.

kahboom (changes requested):

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
stanislavsemeniuk: I used `score.type` as a `key` attribute for mapped components and from what i know `type` is a unique property among scores , so it can't be two elements with similar `type` in one array.

Or it is not what you meant?

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
stanislavsemeniuk: Good catch, fixed it

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
kahboom: Ah yes sorry about that, that's fine

client/src/app/pages/vulnerability-details/vulnerability-details.tsx
stanislavsemeniuk: From what i know , the source should always be "CVE project"

kahboom (approved):
LGTM, thanks @stanislavsemeniuk

PR #991, opened 2026-04-14
feat: CVSS Scores - Vulnerabilities Table page

CryptoRodeo:
JIRA: [TC-4091](https://redhat.atlassian.net/browse/TC-4091)

## Changes

**Tooltip text**:
image

**CVSS Version Badges**:
image

[TC-4091]: https://redhat.atlassian.net/browse/TC-4091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Display CVSS version information alongside vulnerability severity scores in the vulnerabilities table and clarify the CVSS tooltip description.

New Features:
- Show a compact CVSS version badge next to each vulnerability severity score in the vulnerabilities table.

Enhancements:
- Update the severity column tooltip to clarify that CVSS scores come from original CVE Project records and that SBOM view provides product-specific impact.

Tests:
- Add unit tests and supporting mocks for the new CVSS version badge component.

### DIFF
--- a/client/src/app/pages/vulnerability-list/vulnerability-table.tsx
+++ b/client/src/app/pages/vulnerability-list/vulnerability-table.tsx
@@ -11,7 +11,7 @@ import {
   Tr,
 } from "@patternfly/react-table";
 
-import { Flex } from "@patternfly/react-core";
+import { Flex, FlexItem } from "@patternfly/react-core";
 
 import { extendedSeverityFromSeverity } from "@app/api/models";
 import { SeverityShieldAndText } from "@app/components/SeverityShieldAndText";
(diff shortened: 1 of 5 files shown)


client/src/mocks/fileMock.ts
```diff
@@ -0,0 +1 @@
+export default "test-file-stub";
```
carlosthe19916, line 1: Do we need this file?

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
```diff
@@ -115,14 +118,21 @@ export const VulnerabilityTable: React.FC = () => {
...
+                        <SeverityShieldAndText
+                          value={extendedSeverityFromSeverity(
+                            item.base_score?.severity,
+                          )}
+                          score={item.base_score?.score ?? null}
+                          showLabel
+                          showScore
+                        />
+                        <CvssVersionBadge version={item.base_score?.type} />
+                      </Flex>
```
carlosthe19916, line 139: Shouldn't we wrap the items inside Flex using FlexItem?

carlosthe19916 (commented):
Looks great, just minor things below

client/src/app/pages/vulnerability-list/vulnerability-table.tsx
CryptoRodeo: Yes, good catch, wrapped

client/src/mocks/fileMock.ts
CryptoRodeo: I think so, after adding my new test I noticed it kept failing so I looked at the jest config. The jest config already referenced these mock files but they were never actually created: https://github.com/guacsec/trustify-ui/blob/main/client/config/jest.config.ts#L25-L32 

It looks like they weren't needed until now. My test is the first one that renders a PatternFly component (`Label` in `CvssVersionBadge`), and PatternFly internally imports CSS and icons, so Jest needs those stubs to exist. These files fill in what was missing

carlosthe19916: I think we have an opportunity for standardizing the tools we use across projects. In TAS we use Vite and here we just Jest. It is out of scope of this PR and we can do it with more calm and planning.

carlosthe19916 (approved):

client/src/mocks/fileMock.ts
CryptoRodeo: Great idea!

PR #997, opened 2026-04-15
chore: add vercel skills

carlosthe19916:
Vercel skills from https://vercel.com/docs/agent-resources/skills

What do you think guys? installing Skills from https://skills.sh/ comes with a .lock file which also help us to upgrade

### DIFF
--- a/.agents/skills/vercel-composition-patterns/rules/architecture-avoid-boolean-props.md
+++ b/.agents/skills/vercel-composition-patterns/rules/architecture-avoid-boolean-props.md
@@ -1,0 +1,100 @@
+---
+title: Avoid Boolean Prop Proliferation
+impact: CRITICAL
+impactDescription: prevents unmaintainable component variants
+tags: composition, props, architecture
+---
+
+## Avoid Boolean Prop Proliferation
+
+Don't add boolean props like `isThread`, `isEditing`, `isDMThread` to customize
+component behavior. Each boolean doubles possible states and creates
+unmaintainable conditional logic. Use composition instead.
+
+**Incorrect (boolean props create exponential complexity):**
+
+```tsx
+function Composer({
+  onSubmit,
+  isThread,
+  channelId,
+  isDMThread,
+  dmId,
+  isEditing,
+  isForwarding,
+}: Props) {
+  return (
+    <form>
+      <Header />
+      <Input />
+      {isDMThread ? (
+        <AlsoSendToDMField id={dmId} />
+      ) : isThread ? (
+        <AlsoSendToChannelField id={channelId} />
+      ) : null}
+      {isEditing ? (
+        <EditActions />
+      ) : isForwarding ? (
+        <ForwardActions />
+      ) : (
+        <DefaultActions />
+      )}
+      <Footer onSubmit={onSubmit} />
+    </form>
+  )
+}
+```
+
+**Correct (composition eliminates conditionals):**
+
+```tsx
+// Channel composer
+function ChannelComposer() {
+  return (
+    <Composer.Frame>
+      <Composer.Header />
+      <Composer.Input />
+      <Composer.Footer>
+        <Composer.Attachments />
+        <Composer.Formatting />
+        <Composer.Emojis />
+        <Composer.Submit />
+      </Composer.Footer>
+    </Composer.Frame>
+  )
+}
+
+// Thread composer - adds "also send to channel" field
+function ThreadComposer({ channelId }: { channelId: string }) {
+  return (
+    <Composer.Frame>
+      <Composer.Header />
+      <Composer.Input />
+      <AlsoSendToChannelField id={channelId} />
+      <Composer.Footer>
+        <Composer.Formatting />
+        <Composer.Emojis />
+        <Composer.Submit />
+      </Composer.Footer>
+    </Composer.Frame>
+  )
+}
+
+// Edit composer - different footer actions
+function EditComposer() {
+  return (
+    <Composer.Frame>
+      <Composer.Input />
+      <Composer.Footer>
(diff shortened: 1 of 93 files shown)


kahboom (approved):
LGTM!

.agents/skills/vercel-composition-patterns/rules/architecture-avoid-boolean-props.md
```diff
@@ -0,0 +1,100 @@
...
+tags: composition, props, architecture
+---
+
+## Avoid Boolean Prop Proliferation
+
+Don't add boolean props like `isThread`, `isEditing`, `isDMThread` to customize
+component behavior. Each boolean doubles possible states and creates
+unmaintainable conditional logic. Use composition instead.
+
+**Incorrect (boolean props create exponential complexity):**
```
stanislavsemeniuk, line 14: I believe adding set of skills from third parties could lead to the situation where new generated code by ai would not hold the same structure and patterns as already implemented human-generated code in this repo.

F.e this skills that have CRITICAL impact is avoiding boolean props like isEditing, isThread etc and uses composition instead.

I like composition pattern but as i see from a source code we didn't have a single component that uses composition and we have a bunch of components here that have boolean props.

Probably there are more examples like this.

And maybe there is not a bad thing that new code would be better than existing one, would use the best practices etc.

To be honest, i don't have a strong opinion on that just leaving my thoughts here.

carlosthe19916: > I believe adding set of skills from third parties could lead to the situation where new generated code by ai would not hold the same structure and patterns as already implemented human-generated code in this repo.
> 

Yes, this is a good point. After skills like these ones are merged it is important to scan our repository and align it to the new skills so we don't see the scenario that you described.
Locally playing I already scanned the repository against the content of this PR and the result was: https://github.com/guacsec/trustify-ui/issues/999 -> this issue describe the road to align our code to the skills and good practices.


> And maybe there is not a bad thing that new code would be better than existing one, would use the best practices etc.

The old and the new code should match title by title, as  mentioned above, an example of making the old code as good/bad as the new one is https://github.com/guacsec/trustify-ui/issues/999

PhilipCattanach:
@carlosthe19916 Can you hold of with the merge of this PR until we have that chance to discuss it.
Marco is leading the adoption of the Agentic SDLC and he has already developed a number of skills that will enable automation consistently across all repositories that make up RHTPA.
Also Rajan has been analyzing our code base and today will be published an ADR that identifies, flaws, anti-patterns and inconsistencies in the code base.
The intention is update the CONVENTIONS files to reflect how to address these issues then use AI automation to resolve them by applying the conventions.
I'm not saying this PR is the wrong thing to do. But decisions like this shouldn't be made in isolation. And i'd like to know what value these skills are adding that our existing skills don't provide.
Thanks!

carlosthe19916:
> @carlosthe19916 Can you hold of with the merge of this PR until we have that chance to discuss it. Marco is leading the adoption of the Agentic SDLC and he has already developed a number of skills that will enable automation consistently across all repositories that make up RHTPA. Also Rajan has been analyzing our code base and today will be published an ADR that identifies, flaws, anti-patterns and inconsistencies in the code base. The intention is update the CONVENTIONS files to reflect how to address these issues then use AI automation to resolve them by applying the conventions. I'm not saying this PR is the wrong thing to do. But decisions like this shouldn't be made in isolation. And i'd like to know what value these skills are adding that our existing skills don't provide. Thanks!

- @mrrajan Interesting. Please share share with us your code analysis findings in regards of this repository once you are done.
- @mrizzi do you see the content of the PR as a potential blocker for the tools you guys are developing?
- @PhilipCattanach What is the level of Independence the UI team could expect in regards of this repository? I know it is still early but I think it is important to set the rules clear. As a UI team our goals go beyond just TPA so more AI initiatives will come for all the UI repositories we own. I wouldn't like to block any of your guys in any way; likewise, I wouldn't like my team to be blocked so we need to find the right balance so both our teams can be happy.

.agents/skills/vercel-composition-patterns/rules/architecture-avoid-boolean-props.md
kahboom: Just wanted to add a note that just because we add these vercel skills to this repo doesn't mean we would need to keep them. We can decide to swap them out for other skills that better suit us later on.

kahboom:
>Marco is leading the adoption of the Agentic SDLC and he has already developed a number of skills that will enable automation consistently across all repositories that make up RHTPA.

Hi @PhilipCattanach, thanks for bringing this to our attention. We are working on standardizing quality across the UI work we're responsible for, and this was a step towards that.

>And i'd like to know what value these skills are adding that our existing skills don't provide.

Although the specific skills in this PR are a bit more broad, this touches an important point.

By definition, skills are reusable instructions for agents that specialize their abilities. I think the best people to write these specialized instructions, which we will depend on for critical workflows, are the ones who have been doing these workflows manually for some time.

Frontend developers have different problems to solve than backend developers, just as even two separate frontend teams may operate differently based on their customer needs. My point is, how do you know Marco's skills will cover our use cases if only we know what our use cases are?

>I'm not saying this PR is the wrong thing to do. But decisions like this shouldn't be made in isolation.

Totally agree. We've been asked to look into agentic SDLC workflows and, simultaneously, to keep an eye on quality drop. If we're expected to adopt this plugin but not our own stuff, then it's important that we're involved in any workflow standardization that could impact our team. If we don't do that, and we try to use a single tool to cover all use cases, the quality of our products will drop.

gildub:
This seems out of scope, please reopen if necessary.

PR #1002, opened 2026-04-16
fix: avoid bundler to load all icons

carlosthe19916:
###  import { TimesIcon } from "@patternfly/react-icons";

The bundler has to load and parse that barrel file, which pulls in references to all ~1,000 icons. Whether those unused icons end up in the final bundle depends on how well tree-shaking

works — and in practice it's fragile (side effects, CommonJS fallbacks, or bundler limitations can defeat it).

###  import TimesIcon from "@patternfly/react-icons/dist/esm/icons/times-icon";
With the direct path:

The bundler loads exactly one small file containing one icon. There's nothing to tree-shake because nothing extra was imported in the first place.

This PR also aligns the Icon imports to the Examples from the official Patternfly website

## Summary by Sourcery

Enhancements:
- Switch PatternFly icon imports from the package barrel to per-icon module paths to reduce unnecessary bundle loading.

### DIFF
--- a/client/src/app/api/model-utils.ts
+++ b/client/src/app/api/model-utils.ts
@@ -1,14 +1,12 @@
 import type React from "react";
 
 import type { LabelProps, ProgressProps } from "@patternfly/react-core";
-import {
-  SeverityCriticalIcon,
-  SeverityImportantIcon,
-  SeverityMinorIcon,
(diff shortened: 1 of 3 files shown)


CryptoRodeo (approved):
LGTM :shipit:

PR #1008, opened 2026-04-21
feat: add a read-only mode

ctron:
This includes a banner and removing some buttons.

Also see: https://github.com/guacsec/trustify/pull/2337
Also see: TC-4162

## Summary by Sourcery

Introduce a client-wide read-only mode surfaced from server metadata and gate write operations behind it.

New Features:
- Add a ReadOnlyContext backed by a trustify info endpoint to expose the server read-only flag across the React app.
- Display a global banner when the instance is in read-only mode to inform users that modifications are disabled.
- Fetch instance metadata via a new /.well-known/trustify endpoint, wired through client and dev-server proxy configuration.

Enhancements:
- Hide or disable upload, group management, label editing, delete, and importer control actions across SBOM and advisory workflows when in read-only mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ test-results/
 managed_context/
 test_suite_analysis/
 .claude/settings.local.json
+.serena/
+.playwright-mcp/
--- a/client/rsbuild.config.ts
+++ b/client/rsbuild.config.ts
@@ -161,6 +161,10 @@ export default defineConfig({
         target: TRUSTIFICATION_ENV.TRUSTIFY_API_URL || "http://localhost:8080",
         changeOrigin: true,
       },
+      "/.well-known/trustify": {
+        target: TRUSTIFICATION_ENV.TRUSTIFY_API_URL || "http://localhost:8080",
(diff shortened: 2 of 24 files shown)


stanislavsemeniuk (commented):
Thank you @ctron !

The only thing I’m not sure about is using an infinite stale time for the query. There could be a situation where a user opens the UI app and `readOnly` is `false`, but later it gets changed to `true` on the backend. In that case, the UI would never refetch the state , not after 5 or 10 minutes, just never. As a result, the user might start seeing 403 API errors without any indication in the UI, like a banner or disabled tooltips, until they manually refresh the page.

But it feels like this might have been intentional and discussed somewhere , I just couldn’t find any details about it in Jira.

And also, once the backend changes are merged, we’ll need to replace the hardcoded `TrustifyInfo` type with an automatically generated one.

ctron:
Ah yes, claude said it would do it, but then didn't. The idea was that on a 503, the frontend would refresh that state. I'll see it will implement that.

stanislavsemeniuk (approved):

PR #1009, opened 2026-04-21
tests: Replace /api/v3 for /api/v2 endpoints to prevent failures on downstream builds

matejnesuta:
## Summary by Sourcery

Update e2e API sorting tests to target the v2 advisory and vulnerability endpoints and align assertions with the v2 response shape.

Tests:
- Switch vulnerability and advisory sorting tests from /api/v3 to /api/v2 endpoints.
- Adapt vulnerability ID sorting checks to use string-based sorting validation on the identifier field.


ikanias (approved):
LGTM

PR #1023, opened 2026-05-05
feat: set new "total" request parameter

ctron:
In order to request totals for pagination, we must now explicitly request them. This change adds the parameter to the request.

## Summary by Sourcery

Add support for optionally requesting total counts for paginated API responses and align client request/response models with this behavior.

New Features:
- Introduce an optional `total` query parameter on multiple paginated API endpoints to request computation of the total number of matching items.

Enhancements:
- Make paginated result schemas treat the `total` field as optional and nullable instead of always required.
- Update client table-control request parameter builder to send `total=true` when pagination is used so totals are returned where supported.

### DIFF
--- a/client/src/app/hooks/table-controls/getHubRequestParams.ts
+++ b/client/src/app/hooks/table-controls/getHubRequestParams.ts
@@ -72,6 +72,8 @@ interface HubRequestParamsQuery {
    * NOTE: The order of items is defined by the API being called.
    */
   offset?: number;
+  /** Whether to compute and return the total count of matching items. */
+  total?: boolean;
   q?: string;
   sort?: string;
 }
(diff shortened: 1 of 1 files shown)


carlosthe19916 (commented):
@ctron let's change things in order. First https://github.com/guacsec/trustify/pull/2338 needs to be merged, then https://github.com/guacsec/trustify will automatically create a PR with the new openapi spec.

ctron:
> @ctron let's change things in order. First [guacsec/trustify#2338](https://github.com/guacsec/trustify/pull/2338) needs to be merged, then https://github.com/guacsec/trustify will automatically create a PR with the new openapi spec.

The PR is merged. And this PR rebased.

carlosthe19916:
@ctron the tests are failing.
- Also if you don't mind let me take care of this change. This current PR is setting "total=true" to all requests which maybe we could avoid. I'll take a deep look at which endpoints don't need the totals and which ones do, so rather than setting "total=true" to all requests we can set that value depending on the page.

ctron:
Sure. I was trying to speed up things, as right now, the UI is broken.

But I also fully understand my AI approach to this creates code which is … sub-optimal :grimacing:

PR #1030, opened 2026-05-10
fix: Align API calls to include the now optional parameter (total=true|false)

carlosthe19916:
Fixes: https://redhat.atlassian.net/browse/TC-4357

The backend introduced a new query parameter `total` which if passed then the response will include the total count of elements in the rest endpoint requested. By Default the rest api does not return totals. E.g.

```
{"items":[],"total":0}
```

### Main changes:
- The Tanstack Queries does not fetch `total` by default. The caller needs to specify it in `HubRequestParams`
- The LicenseFilters are now rendered always in uppercase. For some reason latest changes of the backend made uppercase/lowercase unstable (sometime it is lowercase, other times it is uppercase) so making it uppercase always secures predictability
- `useGroupForm` and `sbom-group-select.tsx` used to fetch "all items from BD", now it is not possible to fetch all items, the max number of elements is "1000". So setting it to 1000. 
- Now "Empty Tables" does not rely on "totalItemCount" as in the main branch due to a cache introduced in the backend which makes the `total`  unreliable after CREATE/DELETE operations . The issue is briefly described here https://github.com/guacsec/trustify/issues/2355

### Problems

The current implementation of the `total` count on paginated responses generates inconsistent data when CREATE/DELETE operations are done. This generates direct issues on pagination because Pagination navigation depends on exact totals.

After certain operations the UI might end up rendering inconsistent data like `91–100 of 94 items` or:
- Deleting the last item on page 5 but page 5 still “exists”
- Adding an item but pagination still shows old page count
- Empty pages are inconsistent on CRUD operations
- Seems like other tests not related to pagination are also failing (I'm investigating why).

The backend team is aware of this.

## Summary by Sourcery

Ensure tables correctly detect empty states by requesting total counts from backend queries and using both total count and current page items to determine when to show empty results.

Bug Fixes:
- Fix table empty-state detection to rely on both total item count and the absence of current page items to avoid showing blank tables without an empty-state message.

Enhancements:
- Request total item counts in list queries across importers, packages, SBOMs, advisories, SBOM groups, licenses, models, and vulnerabilities to support accurate pagination and empty-state handling.

### DIFF
--- a/client/src/app/Constants.ts
+++ b/client/src/app/Constants.ts
@@ -5,6 +5,8 @@ import ENV from "./env";
 export const FILTER_TEXT_CATEGORY_KEY = "";
 export const FILTER_NULL_VALUE = "\0";
 
+export const MAX_ITEMS_PER_PAGE = 1000;
+
 export const RENDER_DATE_FORMAT = "MMM DD, YYYY";
 export const RENDER_DATETIME_FORMAT = "MMM DD, YYYY | HH:mm:ss";
 export const FILTER_DATE_FORMAT = "YYYY-MM-DD";
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -28,6 +28,7 @@ export interface HubRequestParams {
     pageNumber: number; // 1-indexed
     itemsPerPage: number;
   };
+  total?: boolean;
 }
 
 export interface HubPaginatedResult<T> {
--- a/client/src/app/hooks/table-controls/getHubRequestParams.ts
+++ b/client/src/app/hooks/table-controls/getHubRequestParams.ts
@@ -74,6 +74,7 @@ interface HubRequestParamsQuery {
   offset?: number;
   q?: string;
   sort?: string;
+  total?: boolean;
 }
 
 /**
@@ -108,7 +109,7 @@ export const requestParamsQuery = (
       .join("&");
   }
 
-  return { limit, offset, q, sort };
+  return { limit, offset, q, sort, total: p.total };
 };
 
 export const labelRequestParamsQuery = (labels: Label[] = []) => {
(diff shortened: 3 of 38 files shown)


stanislavsemeniuk:
@carlosthe19916 About this problem:

_The current implementation of the total count on paginated responses generates inconsistent data when CREATE/DELETE operations are done._

According to BE engineer comment in the github issue link you provided, it seems like the problem with "inconsistent totals" would not be solved on a BE side. Am i right or did i miss some discussion?

That's the comment: _It's not a bug, it's a feature. The totals are cached now. Default TTL for the cache is 60s._

carlosthe19916:
@stanislavsemeniuk Yes you are right. The backend engineers considered the new `total` as a feature so they are not planning to change it. 

I documented the problem also in JIRA https://redhat.atlassian.net/browse/TC-4357 so the backend engineers are aware of it together with QE engineers. 

On the UI side, the best we can do is to take the new endpoint and just use it.

stanislavsemeniuk:
@carlosthe19916  Yeah, but there is thing we could do on a fe side.

F.e optimistical decrement of totals after delete operation and optimistical increment of totals after create operation.

Right now in mutate function `onSuccess` uses only `invalidateQueries` that just refetch BE , but we could also try to use `queryClient.setQueriesData` to decrease or increase totals by hand.

Ofc it would not solve the problem completely and kinda weird to be honest , i prefer not to have that in the code, but just in case nothing on BE side would be changed - it feels like we can't just leave it like it is right now

carlosthe19916:
@stanislavsemeniuk that is a good point. The down side is that Create and Delete Operations are not only done by the UI but with direct interaction with the rest endpoints.

- There are CI pipelines (For instance Konflux) that constantly creates SBOMs without using the UI
- The Importers that run in Openshift, similar to cron jobs , are also creating Advisories/Vulnerabilities constantly using internal connections and not using the UI
- There are tools like https://github.com/guacsec/trustify/tree/main/cli that other teams are using for deleting SBOMs (this run both locally and in Kubernetes cron jobs as far as I understand)

So the `queryClient.setQueriesData` would only work for operations triggered by the UI. But the majority of CREATE/DELETE operations are not done through the UI

stanislavsemeniuk (approved):
LGTM! I think we can get back to this problem later in different pr

PR #1055, opened 2026-05-27
chore(eslint): enable no-useless-assignment rule

carlosthe19916:
## Summary
- Remove the `no-useless-assignment` override from ESLint config
- Replace useless initial assignments with `const` + ternary or uninitialized `let` declarations

## Test plan
- [x] `npm run lint` passes with 0 warnings

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by Sourcery

Enable the ESLint no-useless-assignment rule by removing its override and updating code to avoid unnecessary variable reassignments.

Enhancements:
- Simplify conditional value computation by using direct const assignments with ternaries instead of mutable temporaries.
- Remove unused initializations by declaring variables without default values where they are later assigned conditionally.
- Streamline local sorting logic by creating sorted arrays directly instead of reassigning existing variables.
- Improve download verification error handling by returning early on success and preserving the original error as the cause in thrown errors.

Build:
- Update ESLint configuration to stop disabling the no-useless-assignment rule.


carlosthe19916:
Done in main branch

PR #1102, opened 2026-06-18
Atlas Purl Response Tests

vobratil:
A parameterisable test that checks if response is returned for various purls. It is meant to be run against an Atlas instance and as such is skipped by default.

### DIFF
--- a/e2e/tests/api/features/analysis-latest-atlas.ts
+++ b/e2e/tests/api/features/analysis-latest-atlas.ts
@@ -1,0 +1,75 @@
+import { logger } from "../../common/constants";
+import { expect, test } from "../fixtures";
+import { formatTimeElapsed } from "../helpers/general-helpers";
+
+// This is a set of tests that are designed to run againt an Atlas instance, as they assume that certain data is already ingested.
+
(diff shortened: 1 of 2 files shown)


e2e/tests/api/features/analysis-latest-atlas.ts
```diff
@@ -0,0 +1,75 @@
+import { logger } from "../../common/constants";
+import { expect, test } from "../fixtures";
+import { formatTimeElapsed } from "../helpers/general-helpers";
+
+// This is a set of tests that are designed to run againt an Atlas instance, as they assume that certain data is already ingested.
```
mrrajan, line 5:
```suggestion
// This is a set of tests that are designed to run against an Atlas instance, as they assume that certain data is already ingested.
```

mrrajan (approved):
LGTM, except a simple typo. We can take care later too.

e2e/tests/api/features/analysis-latest-atlas.ts
vobratil: @mrrajan Many thanks!

PR #1105, opened 2026-06-22
Fix: Dashboard - Not-disabled changing of Watched SBOMs in ReadOnly mode

stanislavsemeniuk:
## Summary

- Added isDisabled to Select component in ReadOnly mode
- Added informative tooltip

## Related Issues
This pr is trying to resolve this issue: https://redhat.atlassian.net/browse/TC-4818

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation
- [ ] Other (describe below)

## Testing
- [ ] All tests pass (`npm test`)
- [ ] No lint errors (`npm run lint`)
- [ ] No type errors (TypeScript compiles cleanly)
- [ ] New/changed UI behavior visually verified in browser
- [ ] Mock data updated if API types changed

## Screenshots
Screenshot 2026-06-22 at 10 20 42

## Summary by Sourcery

Disable changing of watched SBOMs when the dashboard is in read-only mode and surface a tooltip explaining the restriction.

Bug Fixes:
- Prevent interaction with the watched SBOM selector when the dashboard is in read-only mode to align UI behavior with permissions.

Enhancements:
- Add a contextual tooltip on the watched SBOM selector to explain why it is disabled in read-only mode.

### DIFF
--- a/client/src/app/pages/home/components/WatchedSbom.tsx
+++ b/client/src/app/pages/home/components/WatchedSbom.tsx
@@ -20,16 +20,12 @@ import {
   TextInputGroup,
   TextInputGroupMain,
   TextInputGroupUtilities,
-  Tooltip,
 } from "@patternfly/react-core";
 import TimesIcon from "@patternfly/react-icons/dist/esm/icons/times-icon";
 import text from "@patternfly/react-styles/css/utilities/Text/text";
 
 import { LoadingWrapper } from "@app/components/LoadingWrapper";
-import {
-  READ_ONLY_TOOLTIP,
-  ReadOnlyContext,
-} from "@app/components/ReadOnlyContext";
+import { ReadOnlyContext } from "@app/components/ReadOnlyContext";
 import { useFetchSBOMById, useFetchSBOMs } from "@app/queries/sboms";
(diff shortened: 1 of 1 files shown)


carlosthe19916:
@stanislavsemeniuk LGTM!

Just a general thought:

Adding Tooltips to each button or component that is disabled looks too much. In my humble opinion it should be enough to disable them. The UI has a very visible banner that explains that certain operations are disabled.

I think other read-only components are using Tooltips but I also think they should be limited to being disabled. 

The reason of my comment is just for keeping the code very clean and not customizing things because we are forced to set tooltips. But this is not a blocker just a random idea, WDYT?

stanislavsemeniuk:
@carlosthe19916 to be honest i don't have a strong opinion on that topic. for me both are okay: to have or not to have tooltips

carlosthe19916:
@stanislavsemeniuk then if you don't mind. Please let's keep the code clean and only disable the buttons/dropdown. Let's not add tooltips. It will help us to keep the clean code and reduce the complexity of such things

carlosthe19916 (approved):

PR #1127, opened 2026-07-01
perf: use pre-fetched advisory severity summary on SBOM list

ctron:
## Summary

* Consume the new `?advisories=true` query parameter on the SBOM list endpoint to display severity counts per row
* Eliminates N+1 requests to `GET /sbom/{id}/advisory` (~4s each) that were made per table row

## Dependencies

Depends on backend PR: https://github.com/guacsec/trustify/pull/2437 which adds the `?advisories=true` query parameter to `GET /v3/sbom`.

## Changes

* `SbomVulnerabilities.tsx` - rewritten to accept pre-fetched advisory summary instead of fetching per row
* `sboms.ts` - pass `advisories: true` in the list query
* `sbom-table.tsx` - pass `item.advisories` to the component instead of `item.id`
* `sbom-context.ts` - add `advisories` field to the item type
* `types.gen.ts` - not committed (gitignored), will be regenerated from the updated OpenAPI spec

## Test plan

- [ ] SBOM list page loads severity badges without per-row API calls
- [ ] Severity counts match those from individual `GET /sbom/{id}/advisory` responses
- [ ] Page load time significantly reduced (from ~40s N+1 to single batch query)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Optimize SBOM list vulnerability display by consuming pre-fetched advisory summaries instead of per-row advisory API calls.

New Features:
- Display SBOM vulnerability severities from advisory summaries returned by the SBOM list endpoint.

Enhancements:
- Extend SBOM list items with optional advisory summary data and propagate it through the table context to the vulnerability gallery component.
- Include the advisories flag in SBOM list queries to retrieve aggregated advisory severity data in a single request.

### DIFF
--- a/client/src/app/pages/sbom-list/components/SbomVulnerabilities.tsx
+++ b/client/src/app/pages/sbom-list/components/SbomVulnerabilities.tsx
@@ -1,31 +1,26 @@
 import type React from "react";
 
-import { Skeleton } from "@patternfly/react-core";
-
-import { LoadingWrapper } from "@app/components/LoadingWrapper";
-import { TableCellError } from "@app/components/TableCellError";
+import type { SbomAdvisorySummary } from "@app/client";
 import { VulnerabilityGallery } from "@app/components/VulnerabilityGallery";
-import { useVulnerabilitiesOfSbom } from "@app/hooks/domain-controls/useVulnerabilitiesOfSbom";
+import type { ExtendedSeverity } from "@app/api/models";
 
 interface SBOMVulnerabilitiesProps {
-  sbomId: string;
+  advisories?: SbomAdvisorySummary | null;
 }
 
-export const SBOMVulnerabilities: React.FC<SBOMVulnerabilitiesProps> = ({
-  sbomId,
-}) => {
-  const { data, isFetching, fetchError } = useVulnerabilitiesOfSbom(sbomId);
+const DEFAULT_SEVERITIES: { [key in ExtendedSeverity]: number } = {
+  unknown: 0,
+  none: 0,
+  low: 0,
+  medium: 0,
+  high: 0,
+  critical: 0,
+};
 
-  return (
-    <LoadingWrapper
-      isFetching={isFetching}
-      fetchError={fetchError}
-      isFetchingState={<Skeleton screenreaderText="Loading contents" />}
-      fetchErrorState={(error) => <TableCellError error={error} />}
-    >
-      <VulnerabilityGallery
-        severities={data.summary.vulnerabilityStatus.affected.severities}
-      />
-    </LoadingWrapper>
-  );
+export const SBOMVulnerabilities: React.FC<SBOMVulnerabilitiesProps> = ({
+  advisories,
+}) => {
+  const severities = { ...DEFAULT_SEVERITIES, ...advisories };
+
+  return <VulnerabilityGallery severities={severities} />;
 };
--- a/client/src/app/pages/sbom-list/components/SbomVulnerabilities.tsx
+++ b/client/src/app/pages/sbom-list/components/SbomVulnerabilities.tsx
@@ -1,26 +1,12 @@
 import type React from "react";
 
-import type { SbomAdvisorySummary } from "@app/client";
+import type { RequestedFieldHashMapHashMap } from "@app/client";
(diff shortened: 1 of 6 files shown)


client/src/app/pages/sbom-list/components/SbomVulnerabilities.tsx
```diff
@@ -1,31 +1,26 @@
 import type React from "react";
 
-import { Skeleton } from "@patternfly/react-core";
-
-import { LoadingWrapper } from "@app/components/LoadingWrapper";
-import { TableCellError } from "@app/components/TableCellError";
+import type { SbomAdvisorySummary } from "@app/client";
```
stanislavsemeniuk: I don't see the `SbomAdvisorySummary` type in openapi spec, there are only `SbomAdvisory` (that was used in `getSbomAdvisories` before) or a new type `RequestedField_HashMap_HashMap` (as was added [in this commit](https://github.com/guacsec/trustify-ui/commit/4808564f60c96615f58ebedba79211123342a13d) ).

client/src/app/pages/sbom-list/components/SbomVulnerabilities.tsx
```diff
@@ -1,31 +1,26 @@
...
+  high: 0,
+  critical: 0,
+};
+
 export const SBOMVulnerabilities: React.FC<SBOMVulnerabilitiesProps> = ({
-  sbomId,
+  advisories,
 }) => {
-  const { data, isFetching, fetchError } = useVulnerabilitiesOfSbom(sbomId);
+  const severities = { ...DEFAULT_SEVERITIES, ...advisories };
```
stanislavsemeniuk: I believe the better pattern would be not to add `const DEFAULT_SEVERITIES` where everything is equals to zero but to change `VulnerabilityGalleryProps` from this:

```
interface VulnerabilityGalleryProps {
  severities: { [key in ExtendedSeverity]: number };
}
```

To this:

```
interface VulnerabilityGalleryProps {
  severities: Partial<{ [key in ExtendedSeverity]: number }>;
}
```

This would let us keep SBOMVulnerabilities component clean and just use it like that:

```
export const SBOMVulnerabilities: React.FC<SBOMVulnerabilitiesProps> = ({ advisories }) => (
  <VulnerabilityGallery severities={advisories ?? {}} />
);
```

client/src/app/pages/sbom-list/sbom-context.ts
```diff
@@ -2,7 +2,11 @@ import React from "react";
 
 import type { AxiosError } from "axios";
 
-import type { SbomHead, SourceDocument } from "@app/client";
+import type {
+  SbomAdvisorySummary,
```
stanislavsemeniuk: Same thing here as in the previous comment

_I don't see the SbomAdvisorySummary type in openapi spec, there are only SbomAdvisory (that was used in getSbomAdvisories before) or a new type RequestedField_HashMap_HashMap_

ctron:
@stanislavsemeniuk tried to convince claude to fix the review comments. Maybe you can have another look.

ctron:
@stanislavsemeniuk It still fails, but neither claude nor me have an idea why. We need help.

stanislavsemeniuk:
@ctron The e2e test is failing because it expects `quarkus-bom` SBOM to have total amount of vulnerabilities to be exactly 16. But backend returns 21

```
{
  "medium": 16,
  "high": 4,
  "low": 1
}
```

To be honest i don't know why it expects 16 and why BE returns 21, but if 21 is correct number we can just change '16' to '21' here in `e2e/tests/ui/pages/sbom-list/columns.spec.ts` :

```
await expect(
      table._table
        .locator(`td[data-label="Vulnerabilities"]`)
        .locator("div[aria-label='total']", { hasText: "16" }), //here from 16 to 21
    ).toHaveCount(1);
```

And also about `RequestedFieldHashMapHashMap`. Is the name of this type is intentional? Because for me it feels like some kind of a claude error on a BE side or maybe it is just my lack of knowledge

ctron:
Thanks for helping out with this. Neither claude nor me came to that conclusion 😬 I found an issue in the backend, which explains that discrepancy, I'll be going to fix it and then we can test this again.

carlosthe19916 (approved):

PR #1134, opened 2026-07-02
fix: sbom Search by name instead of full search

carlosthe19916:
## Summary
Fixes https://redhat.atlassian.net/browse/TC-5046

The Search page (/search) was sending SBOM API requests as:

```
  GET /api/v3/sbom?q=~INPUT_TEXT
```

  This performs a full-text search across all fields. As Requested in the jira linked above, for performance reasons we should change it to:

```
  GET /api/v3/sbom?q=name~INPUT_TEXT
```

  What changed

  - Main search submission: When the user types in the search bar and presses Enter, the SBOM endpoint now receives q=name~INPUT_TEXT instead of q=~INPUT_TEXT
  - Autocomplete dropdown: The real-time SBOM suggestions while typing also use q=name~INPUT_TEXT
  - Sidebar filter: The "Filter text" filter is hidden from the left sidebar on the SBOM tab, since the top search bar already handles it
  - Other tabs unaffected: Package, Vulnerability, and Advisory endpoints still use full-text search (q=~INPUT_TEXT) as before

## Related Issues
https://github.com/guacsec/trustify/issues/2440

## Summary by Sourcery

Update SBOM search to query by name instead of full-text and align UI filters with the new search behavior.

Bug Fixes:
- Change SBOM API queries on the search page and autocomplete to use a name-based filter instead of full-text search, addressing performance concerns.

Enhancements:
- Align SBOM table filter state and filter panel configuration with the new name-based search field, including updating filter keys and omitted categories.

### DIFF
--- a/client/src/app/pages/sbom-list/sbom-context.ts
+++ b/client/src/app/pages/sbom-list/sbom-context.ts
@@ -25,7 +25,7 @@ interface ISbomSearchContext {
     | "labels"
     | "vulnerabilities",
     "name" | "published",
-    "" | "published" | "labels" | "license",
+    "name" | "published" | "labels" | "license",
     string
   >;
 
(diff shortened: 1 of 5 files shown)


CryptoRodeo (approved):
LGTM

PR #1149, opened 2026-07-08
test(e2e): add tests for page title updates, upload route, and search button visibility

chetna14manku:
- Add page title verification tests for TC-3370
- Add upload button visibility test for TC-3248
- Add 404 test for /upload route for TC-3218
- Reorganize feature files into @page-title and @search folders

Co-Authored-By: Claude <noreply@anthropic.com>

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation
- [ ] Other (describe below)

## Testing
- [ ] All tests pass (`npm test`)
- [ ] No lint errors (`npm run lint`)
- [ ] No type errors (TypeScript compiles cleanly)
- [ ] New/changed UI behavior visually verified in browser
- [ ] Mock data updated if API types changed

## Summary by Sourcery

Add new BDD end-to-end coverage around recent bug fixes in search and not-found behavior.

Tests:
- Add a BDD scenario to verify /upload correctly shows the 404 Not Found page.
- Add BDD step definitions and scenarios to verify search page title behavior across navigation and tab changes.
- Add BDD step definitions and scenarios to ensure the Upload Advisory button is not shown on the search advisories tab.


e2e/tests/ui/features/@page-title/page-title.step.ts
```diff
@@ -0,0 +1,3 @@
+// Import page-level common steps - no need to define steps here
+// All steps used by @page-title features are defined in page-common-steps.ts
+import "../page-common-steps";
```
vobratil: Not really sure what the idea is behind this file? Seems like it doesn't do anything.

e2e/tests/ui/features/@page-title/verify-page-title.feature
```diff
@@ -0,0 +1,35 @@
...
+
+# Related to TC-3370: Page title does not change based on viewed page
+# The fix ensures page titles update when navigating to different pages
+Scenario: Verify page title changes when navigating to Search page
+	When User navigates to Search results page
+	Then the page title should contain "Search"
+
+# Note: Tabs within the Search page all show "Search" in the title
+# because you're still on the Search page - just viewing different result types
+Scenario: Verify page title remains "Search" when switching between tabs
```
vobratil: I believe that the convention is to have all the `When` conditions in the beginning, followed by all the `Then` expected results. The conditions are `and` conditions. You shouldn't have multiple steps in one scenario. Each of these `When-Then` situations should be a separate scenario.

e2e/tests/ui/features/@page-title/verify-page-title.feature
```diff
@@ -0,0 +1,35 @@
...
+	Then the page title should contain "Search"
+	When Clicks on Advisories tab
+	Then the page title should contain "Search"
+	When User clicks on "SBOMs" tab
+	Then the page title should contain "Search"
+	When User clicks on "Packages" tab
+	Then the page title should contain "Search"
+
+# Verify page titles for main application pages
+Scenario Outline: Verify page title changes when navigating to <page> page
```
vobratil, line 10: This seems redundant with the first scenario. Couldn't just the general "Search" page be included in the list of examples below?

e2e/tests/ui/features/@page-title/verify-page-title.feature
```diff
@@ -0,0 +1,35 @@
...
+Scenario: Verify page title changes when navigating to Search page
+	When User navigates to Search results page
+	Then the page title should contain "Search"
+
+# Note: Tabs within the Search page all show "Search" in the title
+# because you're still on the Search page - just viewing different result types
+Scenario: Verify page title remains "Search" when switching between tabs
+	When User navigates to Search results page
+	Then the page title should contain "Search"
+	When Clicks on Advisories tab
```
vobratil: Missing quotation marks.

e2e/tests/ui/features/@search/verify-no-upload-advisory-button.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Search - Verify No Upload Advisory Button
```
vobratil: This test scenario should probably be lumped together in the file with tests for the search page, not be arbitrarily separated. It's too niche.

e2e/tests/ui/features/page-common-steps.ts
```diff
@@ -0,0 +1,55 @@
+import { createBdd } from "playwright-bdd";
```
vobratil: I'm pretty sure that we don't need any of these steps, as those already exist in other step definition files. Please re-use existing steps instead of duplicating.

vobratil (changes requested):
@chetna14manku I've given it a quick look. I think a necessary change is removing the unneeded `page-title` steps file and the duplicitous step definitions in common steps. We already have these. Otherwise it looks good to me.

chetna14manku:
> @chetna14manku I've given it a quick look. I think a necessary change is removing the unneeded `page-title` steps file and the duplicitous step definitions in common steps. We already have these. Otherwise it looks good to me.

@vobratil I've updated the code changes as per your suggestions. And `@page-title/page-title.step.ts` file is added for the BDD test framework to address changes in TC-3370.
Please review again, thank you!

e2e/tests/ui/features/@page-title/page-title.step.ts
```diff
@@ -0,0 +1,39 @@
+import { createBdd } from "playwright-bdd";
+
+import { test } from "../../fixtures";
+
+import { expect } from "../../assertions";
+
+export const { Given, When, Then } = createBdd(test);
+
+// TC-3370: Page title verification
+When("User navigates to {string} page", async ({ page }, pageName: string) => {
```
vobratil: I thought that we had this step already implemented, but seems like I was wrong in my previous review. Apologies for that.

However, we have this particular action [implemented as a method](https://github.com/guacsec/trustify-ui/blob/5df79a2d7d71895ad753ad5bb96b42b764f2360d/e2e/tests/ui/pages/Navigation.ts#L17), which we should re-use in order to prevent code redundancy.

e2e/tests/ui/features/@page-title/page-title.step.ts
```diff
@@ -0,0 +1,39 @@
...
+    Vulnerabilities: "/vulnerabilities",
+    Packages: "/packages",
+    Importers: "/importers",
+  };
+  const route = pageRoutes[pageName];
+  if (!route) {
+    throw new Error(`Unknown page: ${pageName}`);
+  }
+  await page.goto(route);
+  await page.waitForLoadState("networkidle");
```
vobratil: I'm not really sure about this condition. Generally, we don't wait for the server to settle down after going to a different page and I'm afraid that this could introduce additional issues. The tests are usually executed using parallel workers and I think that they may mess with this expectation. But even if they don't, the test steps have default timeout period, so it's unnecessary. I would like to ask for its removal.

e2e/tests/ui/features/@page-title/page-title.step.ts
```diff
@@ -0,0 +1,39 @@
...
+});
+
+When("User clicks on {string} tab", async ({ page }, tabName: string) => {
+  const tab = page.getByRole("tab", { name: tabName });
+  await tab.click();
+  await page.waitForLoadState("networkidle");
```
vobratil: Same as above.

e2e/tests/ui/features/@search/search.step.ts
```diff
@@ -0,0 +1,28 @@
...
+import { test } from "../../fixtures";
+
+import { expect } from "../../assertions";
+
+export const { Given, When, Then } = createBdd(test);
+
+// Navigation steps for search feature
+When("User navigates to Search results page", async ({ page }) => {
+  await page.goto("/search");
+  await page.waitForLoadState("networkidle");
```
vobratil: Please see my other comment. You don't need to define this step, because you are already defining it in `e2e/tests/ui/features/@page-title/page-title.step.ts`. Please re-use it.

e2e/tests/ui/features/@search/search.step.ts
```diff
@@ -0,0 +1,28 @@
...
+});
+
+When("Clicks on Advisories tab", async ({ page }) => {
```
vobratil: Same as above.

e2e/tests/ui/features/@search/search.step.ts
```diff
@@ -0,0 +1,28 @@
...
+  const advisoriesTab = page.getByRole("tab", { name: "Advisories" });
+  await advisoriesTab.click();
+  await page.waitForLoadState("networkidle");
```
vobratil: See my comment about this condition in the other file.

vobratil (changes requested):
@chetna14manku I've given it another glance. I'm afraid that there is still some important redundant code. Please de-duplicate the methods as per my review. Otherwise the PR looks good.

e2e/tests/ui/features/@page-title/page-title.step.ts
chetna14manku: Hi @vobratil , Added `steps/navigation.ts` by importing the existing Navigation class from  `pages/Navigation.ts` to define shared BDD steps needed by `@page-title` and `@search`. This avoids duplication and is reusable by other features. 
Also removed all waitForLoadState("networkidle") calls as suggested.
Thank you!

vobratil (approved):
@chetna14manku Okay, I think that the PR is looking good now. I've triggered the CI checks. If everything passes, we can merge the changes :).

PR #1156, opened 2026-07-13
chore: use tsconfig.app.json in client ts check

carlosthe19916:
## Summary
- `client/package.json` now uses `-p tsconfig.app.json` otherwise the typescript checks in the client won't be triggered
- After enabling the option above some typescript checks were raised hence fixing them in this PR.

## Summary by Sourcery

Strengthen client-side TypeScript linting by targeting tsconfig.app.json and fix resulting type and layout issues across advisory flows, upload/read-only handling, ref initialization, and home page layout.

Bug Fixes:
- Align advisory deletion flows and success messaging with the AdvisoryHead type to satisfy stricter TypeScript checks.
- Initialize React ref hooks with explicit null defaults to resolve TypeScript strictness issues.
- Update read-only handling in upload and list pages to use the areMutationsDisabled flag instead of removed helpers and props.
- Correct layout props in the home attention section to use the flexWrap prop expected by the Flex component.

Enhancements:
- Tighten client linting by running TypeScript checks against tsconfig.app.json.
- Simplify ReadOnlyContext usage across SBOM and importer list components by removing unused tooltip and helper imports.

Build:
- Update the client lint script to run tsc with tsconfig.app.json for type checking.

### DIFF
--- a/client/src/app/pages/advisory-details/advisory-details.tsx
+++ b/client/src/app/pages/advisory-details/advisory-details.tsx
@@ -33,7 +33,7 @@ import {
 } from "@app/Constants";
 import { PathParam, Paths, useRouteParams } from "@app/Routes";
 import { ExtendedSeverity } from "@app/api/models";
-import { type AdvisorySummary } from "@app/client";
+import { AdvisoryHead } from "@app/client";
 import { ConfirmDialog } from "@app/components/ConfirmDialog";
 import { DocumentMetadata } from "@app/components/DocumentMetadata";
 import { LoadingWrapper } from "@app/components/LoadingWrapper";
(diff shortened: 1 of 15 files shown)


client/src/app/pages/advisory-details/advisory-details.tsx
```diff
@@ -33,7 +33,7 @@ import {
 } from "@app/Constants";
 import { PathParam, Paths, useRouteParams } from "@app/Routes";
 import { ExtendedSeverity } from "@app/api/models";
-import { type AdvisorySummary } from "@app/client";
+import { AdvisoryHead } from "@app/client";
```
stanislavsemeniuk, line 36: Can we also import it as a `type`?

stanislavsemeniuk: Sorry, we can forget about it, i thought we have `@typescript-eslint/consistent-type-imports` rule in config but we don't so i guess it is not important

stanislavsemeniuk (approved):

client/src/app/pages/advisory-details/advisory-details.tsx
carlosthe19916: oh, I didn't see your comment. Sorry, but that is a good idea. We can enable that rule, it would add value

